PR #1, opened 2019-07-10
add naive match

kmelve:
This adds a naive implementation of the `match` operator.


judofyr:
Looks good to me so far. We need to setup a proper test suite before we can implement these properly anyway.

PR #5, opened 2019-08-16
update sample in readme so it outputs data

billymoon:


kmelve:
Good catch - thanks!

judofyr:
You might also be interested in following #4 which is where I'm currently doing active development.

PR #10, opened 2019-10-14
Add error handler to parser

asbjornh:
*What:* Adds an `errorHandler` (function) argument to `parse`, which is called with the result object when there's an error. The default value for `errorHandler` is a function that throws an error.

*Why:* I want to use this for an eslint plugin to validate tagged groq queries in javascript, but it turns out that the error message `"Syntax error in GROQ query"` is not that helpful for large queries. The raw parser is more helpful (returning a `position`), but as far as I can tell the raw parser is not a part of the public API, so I'm not as comfortable using that directly. An error handler that exposes the `result` object from the raw parser solves this.

Here's an image for a bit more context (spot the typo!):

bilde

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,7 +1,7 @@
 # API Reference
 * [groq-js](#module_groq-js)
     * [.evaluate(tree, [options])](#module_groq-js.evaluate) ⇒ [<code>Value</code>](#Value)
-    * [.parse(input)](#module_groq-js.parse) ⇒ [<code>SyntaxNode</code>](#SyntaxNode)
+    * [.parse(input, [errorHandler])](#module_groq-js.parse) ⇒ [<code>SyntaxNode</code>](#SyntaxNode)
 
 
 <a name="module_groq-js.evaluate"></a>
@@ -21,7 +21,7 @@ Evaluates a syntax tree (which you can get from [parse](#module_groq-js.parse)).
 
 <a name="module_groq-js.parse"></a>
 
-### groq-js.parse(input) ⇒ [<code>SyntaxNode</code>](#SyntaxNode)
+### groq-js.parse(input, [errorHandler]) ⇒ [<code>SyntaxNode</code>](#SyntaxNode)
 Parses a GROQ query and returns a tree structure.
 
 **Kind**: static method of [<code>groq-js</code>](#module_groq-js)  
@@ -29,6 +29,7 @@ Parses a GROQ query and returns a tree structure.
 | Param | Type | Description |
 | --- | --- | --- |
 | input | <code>string</code> | GROQ query |
+| errorHandler | <code>function</code> | Function to handler parsing errors |
 
 
 <a name="Value"></a>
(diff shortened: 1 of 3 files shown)


judofyr:
Yeah, I'd rather not expose too many details of the raw parser since it will probably be tweaked and changed in the future. The error reporting is indeed very lacking at the moment, and I have it on my TODO list to improve it.

One thing we could do is to to expose a `position`-property on the error object. Then you would be able to do:

```javascript
try {
  parse(…)
} catch (err) {
  console.log("Error at: " + err.position);
}
```
Would this work for you?

asbjornh:
Oh, that makes sense. Exposing it on the error would also work fine! I’ll make an update tomorrow :)

judofyr:
I'm also not quite sure what's the best way to have custom errors in JavaScript. Should we subclass it? `class GROQSyntaxError extends Error { … }`?

asbjornh:
That seems to be the way people do it, yeah! :) I'll give it a shot

judofyr:
Looks good! How about a test in `test/parser.test.js` as well? :)

asbjornh:
I updated it now. I took the liberty of adding the position to the error message. The wording might be a bit misleading now though, since I'm guessing that the reported position is pointing to an unexpected token instead of the actual syntax error, so maybe change the wording to `Unexpected token at position x` or just use the old wording.

judofyr:
I think the error message is good. In the future we probably want to add line/column, but let's start with this.

asbjornh:
> I think the error message is good. In the future we probably want to add line/column, but let's start with this.

That would be awesome! I'll have to derive line/col manually for my eslint plugin for now but the real values would be even better :)

> Looks good! How about a test in test/parser.test.js as well? :)

Will do!

judofyr:
> That would be awesome! I'll have to derive line/col manually for my eslint plugin for now but the real values would be even better :)

The plan was for the AST to follow the `ESTree` node structure: https://github.com/estree/estree/blob/master/es5.md#node-objects. This involves tracking line/column for all the AST nodes and while doing that work it should be easy to emit line/column for errors as well.

asbjornh:
Thanks! :)

judofyr:
I've released v0.1.4 now!

asbjornh:
Sweet! My brain now feels much better about this:

bilde

judofyr:
Nice! Is this plugin open source?

asbjornh:
For now it's on my machine :) I thought about publishing it but as you probably know eslint plugins are tied to their npm package names, so I don't really want to steal the `eslint-plugin-sanity` or `eslint-plugin-groq` names so I'm just running it through `eslint-plugin-local`. If you'd like to host  an eslint plugin I'd be happy to submit a PR

asbjornh:
> Nice! Is this plugin open source?

@judofyr It is now 🙂 
https://github.com/asbjornh/eslint-plugin-groq
https://www.npmjs.com/package/@asbjorn/eslint-plugin-groq

judofyr:
This is very cool! We'll look into creating an official plugin later.

PR #17, opened 2020-03-22
Migrate to TypeScript

rexxars:
Let's discuss whether or not this is something we want to do, and which changes needs to be in place!


rexxars:
> What's up with the additional tests? Most of the ones you added seems to be covered by https://github.com/sanity-io/groq-test-suite.

I was trying to understand the codebase and wrote some tests while doing so. Removed the `references()` tests.

> Why turning `rawParser.js` into a `.ts` file? Why not keep it and just add a `.d.ts` file?
Good point. Fixed.

> Any significant difference in bundle size?
I don't think we've set a browser support level for the current version and we're not generating any babelified + minified version, so hard to tell. According to bundlephobia, it's currently 12.5kB after minification + gzipping - I'm assuming babelifying it will add quite a bit to that.

The typescript version is 10kB after gzipping, and that's with the target set to ES5.

I'll squash-merge the commits if you're happy.

rexxars:
The tsdoc currently used generates markdown - we could potentially include it in the repo if you want? An alternative is to generate HTML (the default for tsdoc) and host it on something like github pages?

I kind of wish the generated docs were a little more readable, but I guess _some_ documentation is better than none?

judofyr:
Replaced by #18

PR #25, opened 2020-11-10
Implement string function

israelroldan:
There's a long standing issue (https://github.com/sanity-io/sanity/issues/1108) for this function and we recently bumped into an query that would benefit greatly of this type coercion possibility.

I took a stab at it and here's my proposal, the `string()` function takes one argument that can be an expression or a static value as with pretty much all other functions.

If the argument is a `boolean`, `number` or `string` the casting is pretty straightforward, for `array` the value is joined with a comma (`,`) and objects are `JSON.stringify`-ed.

I'd love to hear what you guys think about this!

### DIFF
--- a/src/evaluator/functions.ts
+++ b/src/evaluator/functions.ts
@@ -139,6 +139,23 @@ functions.select = async function select(args, scope, execute) {
   return NULL_VALUE
 }
 
+functions.string = async function string(args, scope, execute) {
+  let value = await execute(args[0], scope)
+  switch (value.getType()) {
+    case 'array':
+      return new StaticValue((await value.get()).join(','));
+    case 'object':
+      return new StaticValue(JSON.stringify(await value.get()));
+    case 'number':
+    case 'string':
+    case 'boolean':
+      return new StaticValue('' + await value.get());
+    default:
+      return NULL_VALUE;
+  }
+}
+functions.string.arity = 1
+
 functions.references = async function references(args, scope, execute) {
   let idValue = await execute(args[0], scope)
   if (idValue.getType() !== 'string') return FALSE_VALUE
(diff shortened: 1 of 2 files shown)


judofyr:
Very cool work!

I'm not sure about the `array` and `object` behavior:

- For arrays I can see that people _might_ end up writing queries like `*[_type == "foo" && string(data) == "[1, 2]"]` (since we don't have a deep-equality operator yet). This is going to be tricky to optimize in Sanity's data store so we'd like to discourage it (until we get proper deep-equality implemented).
- You probably want to be able to join by something else than `, ` as well.
- It seems a bit strange that JSONification is only available for `object` (and not `array`).

What about the following changes in this PR:

- Make `string(arr)` and `string(obj)` return `null`

And then we can tackle the rest with new functions (in a new PR):

- Introduce a `join(arr, delim)` function.
- Introduce a `json_encode(any)` function (potentially also with a `json_decode(str)`?)

Note: We'll also have to write up specs for this (in http://github.com/sanity-io/GROQ).

israelroldan:
Thanks for the great feedback @judofyr! 
Indeed I thought that string comparison of objects was a 50/50 "may be a good thing, may not" hehe.

For now I have removed the `array` and `object` behavior so they now evaluate to `null`.
I'm all up for those new functions (in a new PR, will give it a try whenever I have some free cycles).

For the spec: https://github.com/sanity-io/GROQ/pull/38

judofyr (approved):

judofyr:
I've rebased it, fixed one test case (due to change in behavior of `null` in objects since you wrote this PR) and merged it!

Thanks for the contribution 🎉

israelroldan:
wohooo that's awesome! :)

PR #26, opened 2020-11-18
Switch to GitHub Actions for running tests

judofyr:
I also ended up adding Node 10.x to the test matrix and discovered that the ordering function wasn't correctly implemented.


rexxars (approved):
👍  Nice

PR #32, opened 2021-01-12
Implement new array traversal semantics

judofyr:
Implements https://github.com/sanity-io/GROQ/issues/42. Tests are intentionally not passing since groq-test-suite is not updated.


judofyr:
Superseded by #35

PR #33, opened 2021-01-13
Implement dateTime

judofyr:
Currently this does not round-trip the time offset (which currently is enforced by the spec).


rexxars (changes requested):
An unused import, but otherwise this looks good.

One question (as I mentioned in [this comment](https://github.com/sanity-io/GROQ/issues/45#issuecomment-760475125)) is whether or not we actually need to use date-fns if all the inputs are RFC 3339.

I think we should be able to parse it using `new Date(input)` and check for invalid dates using `isNaN(date.getTime())`, but I suppose we would also be accepting non-RFC3339 strings in this case. Maybe a regex would be enough to catch that, though?

judofyr:
I removed the dependency on date-fns and tweaked it to match the updated semantic.

rexxars (approved):
👍

PR #36, opened 2021-03-09
Use standardized ESLint

judofyr:
I've disabled a few lints here and there. In the future we can work on refactoring the code to remove these.


rexxars (approved):
👍

PR #40, opened 2021-06-17
feat: compare-util

ricokahler:
The following is a work-in-progress version of the compare util. I tried a few faster ways to get this working but this is the only version I have that passes all my tests…

~it runs in `O(2^n)` 😭~ I figured out a way to speed this up significantly by removing states where expressions with certain equalities can't both be true at the same time (e.g. the expressions `_type == 'movie'` and `_type == 'book'` can't both be true at the same time)

There's still a decent amount of work left for this but there is a path forward now given the speed up. What's left:

- [x] expressions with inequalities e.g. `x > 5 && x < 10`
- [ ] expressions with array membership
- [ ] expressions with filters
- [ ] expressions with certain functions


src/utils/compare.ts
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  findHashes(treeA)
+  findHashes(treeB)
+
+  let m: {[hash: string]: boolean}[] = []
+
+  const sorted = Array.from(hashes).sort((a, b) => a.localeCompare(b, 'en'))
+
+  // 2^n… oof
```
judofyr, line 212: 📈

ricokahler:
i'll revisit this one day maybe 😅

PR #44, opened 2021-08-04
Support specifying identity/now

judofyr:


sgulseth (commented):
Code LGTM! Could be nice to have some tests to ensure context is handled when passed to children?

judofyr:
> Code LGTM! Could be nice to have some tests to ensure context is handled when passed to children?

I added some tests — which should also cover the cases where the context is passed to a new scope.

sgulseth (approved):
Nice, LGTM! :)

PR #52, opened 2022-03-18
Parse object expressions starting with string

judofyr:
Previously when we saw `{"foo"` we would immediately start parsing it as a field (`{"foo":123}`). This is not correct since it's possible to write `{"foo" == bar => {…}}`). This changes the logic to always parse it as a general expression and if it ends up being parsed as a literal string _then_ we invoke the logic for field.


dcilke (approved):
Looks good!

PR #57, opened 2022-06-15
Implement experimental string and array functions

atombender:
Requires [this test suite PR](https://github.com/sanity-io/groq-test-suite/pull/52).


src/evaluator/functions.ts
```diff
@@ -279,6 +301,57 @@ string.startsWith = async function (args, scope, execute) {
...
+  for await (const elem of arr) {
+    const val = await elem.get()
+    if (val !== null) {
+      result.push(val)
+    }
+  }
+
+  return fromJS(result)
+}
+array.compact.arity = 0
```
judofyr, line 353:
```suggestion
array.compact.arity = 1
```

atombender:
Realized none of the functions are being tested, because `experimentalFunctions` is not in the supported feature list. Will work on this.

PR #59, opened 2022-06-16
feat: add unique method on the array namespace

sgulseth:


test/functions.test.ts
```diff
@@ -29,4 +29,63 @@ t.test('Functions', async (t) => {
       t.same(data.deep.time, data.topTime)
     })
   })
+
+  t.test('Array namespace', async (t) => {
```
atombender, line 33: Isn't the function already covered by the GROQ test suite?

atombender (changes requested):

src/evaluator/functions.ts
```diff
@@ -441,11 +441,44 @@ diff.changedOnly = () => {
...
+      case 'boolean':
+      case 'datetime':
+        if (result.includes(element) === false) {
+          result.push(element)
+        }
+        break
+      default:
+        result.push(element)
+    }
+  }
```
atombender, line 471: Avoid linear search?

```suggestion
  const result = []
  const added = new Set()
  for await (const iter of value) {
    const element = await iter.get()
    switch (iter.type) {
      case 'number':
      case 'string':
      case 'boolean':
      case 'datetime':
        if (!added.has(element.data)) {
          added.add(element.data)
          result.push(element)
        }
        break
      default:
        result.push(element)
    }
  }
```

sgulseth: Perf! Good point

test/functions.test.ts
sgulseth: Not atm, and I wanted something to test with locally. Can add some tests to the test suite though!

atombender: We have tests [here](https://github.com/sanity-io/groq-test-suite/pull/54).

sgulseth: Ah! Removed, tested and verified with `GROQTEST_SUITE=../groq-test-suite/suite.ndjson ./test/generate.sh` from the branch

atombender:
Great. Awaiting review on [spec PR](https://github.com/sanity-io/GROQ/pull/82) from @judofyr.

src/evaluator/functions.ts
```diff
@@ -456,11 +456,47 @@ diff.changedOnly = () => {
 }
 diff.changedOnly.arity = 3
 
+const array: FunctionSet = {}
+
+array.unique = async function (args, scope, execute) {
+  if (args.length > 1) {
+    throw new Error('Incorrect number of arguments to function "array::unique"')
+  }
```
judofyr, line 464: This should be handled by setting `array.unique.arity = 1` instead. That will correctly treat `array::unique(foo, bar)` as a compile-time error.

src/evaluator/functions.ts
```diff
@@ -456,11 +456,47 @@ diff.changedOnly = () => {
...
+          result.push(iter.data)
+        }
+        break
+      default:
+        const element = await iter.get()
+        result.push(element)
+    }
+  }
+
+  return fromJS(result)
```
judofyr, line 491: It's possible to implement this in a streaming manner:

```suggestion
  return new new StreamValue(async function* () {
    const added = new Set()
    for await (const iter of value) {
      switch (iter.type) {
        case 'number':
        case 'string':
        case 'boolean':
        case 'datetime':
          if (!added.has(iter.data)) {
            added.add(iter.data)
            yield iter
          }
          break
        default:
          yield iter
      }
    }
  })
```

judofyr (changes requested):

judofyr (approved):

PR #62, opened 2022-07-01
feat: implement selector validation

codebymatt:
## The Problem
[The GROQ spec](https://sanity-io.github.io/GROQ/draft/#sec-Selector) requires support for selector validation. This has yet to be added to `groq-js`. Selector validation is a pre-requisite to supporting the `diff` extension in `groq-js`.

## Changes
* Add `SELECTOR_BUILDER` to the parser to handle validation and building of selectors
* Update `rawParser` to allow tuples to valid traversal elements
* Add step in `rawParser` to transform traversals in the form `thing.(foo, bar)` to `(thing.foo, thing.bar)` (main implementation is in `helpers/parserHelper.ts`)
* Add unit tests to verify the basic implementation works

## Notes
I'm not convinced transforming `traversals with tuples` into `tuples containing traversals` is the optimum solution, but it's the least hacky way I can see of tackling this with the context I currently have.

I suspect a better solution is to rework how we parse traversals in general, but that part of the code is still a bit of black box at the moment. I'm also hesitant to add new syntax to the general parsing code since traversals with tuples is only intended for use in selectors. I'll keep chewing on the problem; if I can think of a better solution I'll revisit this.

## Todo
- [x] Add support for selectors in the form `doc.(foo, bar)`
- [ ] ~~Ensure there are tests in `groq-test-suite` covering selectors, and if not, add them~~
  I'm going to create a new shortcut ticket for this since the test suite lives in a different repo


src/parser.ts
```diff
@@ -696,6 +760,10 @@ function validateArity(name: string, arity: GroqFunctionArity, count: number) {
   }
 }
 
+function argumentShouldBeSelector(namespace: string, args: NodeTypes.ExprNode[]) {
+  return namespace == 'diff' && AbortSignal.length == 2
```
atombender, line 764: `AbortSignal` should be `args`, right?

src/markProcessor.ts
```diff
@@ -82,6 +82,10 @@ export class MarkProcessor {
     this.index += 1
   }
 
+  unshift(): void {
+    this.index -= 1
```
atombender, line 86: I'd add a sanity check here and throw an error if `this.index === 0`.

src/parser.ts
codebymatt: Yes absolutely, don't know how that got in there.

src/parser.ts
```diff
@@ -662,6 +672,60 @@ const TRAVERSE_BUILDER: MarkVisitor<(rhs: TraversalResult | null) => TraversalRe
   },
 }
 
+const SELECTOR_BUILDER: MarkVisitor<NodeTypes.SelectorNode> = {
+  this_attr(p) {
+    const name = p.processString()
+
+    if (['null', 'true', 'false'].includes(name)) throw new GroqSelectorError('invalid selector')
```
atombender, line 679: Isn't this a bit naive, and ignores other invalid things like numbers? Should we not instead check that it is a valid identifier?

src/parser.ts
```diff
@@ -696,6 +760,10 @@ function validateArity(name: string, arity: GroqFunctionArity, count: number) {
   }
 }
 
+function argumentShouldBeSelector(namespace: string, args: NodeTypes.ExprNode[]) {
+  return namespace == 'diff' && args.length == 2
```
atombender, line 764: I think this needs to be more specific about the function names it tests for. Also, `diff::changedAny()` and `diff::changedOnly()` take three arguments, not two, and only the last argument is a selector.

atombender (changes requested):
Aside from the comments, this looks reasonable, though I think we need @judofyr to also review, as I'm not very familiar with this. One thing, I'm not sure why we need to identify selectors during parsing. Selectors should only come into play when evaluating.

codebymatt:
> I'm not sure why we need to identify selectors during parsing. Selectors should only come into play when evaluating.

That's a fair point. As far as I can tell we don't use the syntax `foo.(a, b)` anywhere else, but are introducing it with selectors (this is the part where I may be wrong).

We need to be able to parse that sort of syntax, but I don't know if we want to allow it anywhere else besides selectors.

Holm also suggested hinted at adding a new builder (in this case `SELECTOR_BUILDER`) to handle validation, which led me to handling the validation during parsing.

I'm starting to wonder if it's worth shelving most of my `groq-js work` until Holm is back.

atombender:
Not sure. I'm not familiar enough with groq-js to say for sure. I did see Holm's comment about adding a new mark processor for selectors, but maybe he meant tuples?

In go-groq, the parser doesn't validate — tuples are just normal expressions that get parsed into the AST, and so they can be used anywhere, in theory. The validation happens during query planning.

src/parser.ts
```diff
@@ -696,6 +776,78 @@ function validateArity(name: string, arity: GroqFunctionArity, count: number) {
...
+  namespace: string,
+  functionName: string,
+  args: NodeTypes.ExprNode[]
+) {
+  return (
+    namespace == 'diff' && args.length == 2 && ['changedAny', 'changedOnly'].includes(functionName)
+  )
+}
+
+// An array of arrays where each internal array consists of traversals to be applied to a base node.
```
codebymatt, line 789: The rest of this file is littered with comments. I generally prefer avoiding comments with well named functions/variables and properly broken down functions, but since this code is so abstract (and hard to follow without context) I've attempted to provide explanation of why we need to do things, and in some cases, what we're actually doing.

judofyr:
Just to clarify:

> One thing, I'm not sure why we need to identify selectors during parsing. Selectors should only come into play when evaluating.

groq-js's "Parser" is a bit misnamed. The actual steps in groq-js is:

- `rawParser`: This takes a plain string and produces the "marks list". This is a "traditional" parser and handles tokenization, operator precedence etc. The marks list is conceptually a tree structured as prefix notation.
- `Parser`: This takes the marks list and produces a more classical tree-structure (`ExprNode`) designed for execution. This does a quite a few things: It handles array traversals, validates functions (since this is a compile-time validation in GROQ), and disambiguates square brackets. Selectors are supposed to be validated at compile-time and hence it needs to be handled in `Parser`.

codebymatt:
After running through the requirements with Holm I'm closing this PR in favour of [a new one](https://github.com/sanity-io/groq-js/pull/63).

PR #63, opened 2022-07-19
feat: add validation for selector syntax

codebymatt:
[The GROQ spec](https://sanity-io.github.io/GROQ/draft/#sec-Selector) requires support for selector validation. This has yet to be added to `groq-js`. Selector validation is a pre-requisite to supporting the `diff` and `delta` extensions in `groq-js`.

Note that full selector syntax isn't available yet, we need to add support for tuples and groups when parsing traversals.

# Changes
* Add a `SELECTOR_BUILDER` to the parser to handle selector validation
* Add relevant unit tests

### DIFF
--- a/src/markProcessor.ts
+++ b/src/markProcessor.ts
@@ -82,6 +82,10 @@ export class MarkProcessor {
     this.index += 1
   }
 
+  unshift(): void {
+    this.index -= 1
+  }
+
   process<T>(visitor: MarkVisitor<T>): T {
     const mark = this.marks[this.index]
     this.shift()
--- a/src/markProcessor.ts
+++ b/src/markProcessor.ts
@@ -82,10 +82,6 @@ export class MarkProcessor {
     this.index += 1
   }
 
-  unshift(): void {
-    this.index -= 1
-  }
-
   process<T>(visitor: MarkVisitor<T>): T {
     const mark = this.marks[this.index]
     this.shift()
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -355,14 +355,14 @@ const EXPR_BUILDER: MarkVisitor<NodeTypes.ExprNode> = {
       return result
     }
 
-    let args: NodeTypes.ExprNode[] = []
+    const args: NodeTypes.ExprNode[] = []
 
     while (p.getMark().name !== 'func_args_end') {
-      if (argumentShouldBeSelector(namespace, name, args)) {
+      if (argumentShouldBeSelector(namespace, name, args.length)) {
         // Since the diff/delta functions aren't validated yet we only want to validate the selector
         // being used. We expect the null valued arg to throw an error at evaluation time.
         p.process(SELECTOR_BUILDER)
-        args.push(NodeTypes.SelectorNode)
+        args.push({type: 'Selector'})
         break
       } else {
         args.push(p.process(EXPR_BUILDER))
@@ -673,19 +673,9 @@ const TRAVERSE_BUILDER: MarkVisitor<(rhs: TraversalResult | null) => TraversalRe
 }
 
 const SELECTOR_BUILDER: MarkVisitor<null> = {
-  square_bracket(p) {
-    p.process(TRAVERSE_BUILDER)
-    return null
-  },
-
-  slice(p) {
-    p.process(TRAVERSE_BUILDER)
-    return null
-  },
-
   group(p) {
-    // This should only throw an error until we add support for groups in selectors.
-    throw new Error('Invalid selector syntax')
+    p.process(SELECTOR_BUILDER)
+    return null
   },
 
   everything() {
@@ -705,18 +695,16 @@ const SELECTOR_BUILDER: MarkVisitor<null> = {
   },
 
   traverse(p) {
-    // p.unshift()
-    const _ = p.process(EXPR_BUILDER)
-    p.process(TRAVERSE_BUILDER)
+    p.process(SELECTOR_BUILDER)
+    while (p.getMark().name !== 'traversal_end') {
+      p.process(TRAVERSE_BUILDER)
+    }
+
+    p.shift()
     return null
   },
 
   this_attr(p) {
-    p.processString()
-    return null
-  },
-
-  attr_access(p) {
     p.processString()
     return null
   },
@@ -865,16 +853,10 @@ function validateArity(name: string, arity: GroqFunctionArity, count: number) {
   }
 }
 
-function argumentShouldBeSelector(
-  namespace: string,
-  functionName: string,
-  args: NodeTypes.ExprNode[]
-) {
+function argumentShouldBeSelector(namespace: string, functionName: string, argCount: number) {
   const functionsRequiringSelectors = ['changedAny', 'changedOnly']
 
-  return (
-    namespace == 'diff' && args.length == 2 && functionsRequiringSelectors.includes(functionName)
-  )
+  return namespace == 'diff' && argCount == 2 && functionsRequiringSelectors.includes(functionName)
 }
 
 class GroqSyntaxError extends Error {
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -355,10 +355,20 @@ const EXPR_BUILDER: MarkVisitor<NodeTypes.ExprNode> = {
       return result
     }
 
-    const args: NodeTypes.ExprNode[] = []
+    let args: NodeTypes.ExprNode[] = []
+
     while (p.getMark().name !== 'func_args_end') {
-      args.push(p.process(EXPR_BUILDER))
+      if (argumentShouldBeSelector(namespace, name, args)) {
+        // Since the diff/delta functions aren't validated yet we only want to validate the selector
+        // being used. We expect the null valued arg to throw an error at evaluation time.
+        p.process(SELECTOR_BUILDER)
+        args.push(NodeTypes.SelectorNode)
+        break
+      } else {
+        args.push(p.process(EXPR_BUILDER))
+      }
     }
+
     p.shift()
 
     if (namespace === 'global' && (name === 'before' || name === 'after')) {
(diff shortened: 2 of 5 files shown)


src/parser.ts
```diff
@@ -662,6 +672,165 @@ const TRAVERSE_BUILDER: MarkVisitor<(rhs: TraversalResult | null) => TraversalRe
   },
 }
 
+const SELECTOR_BUILDER: MarkVisitor<null> = {
```
codebymatt, line 675: Is it worth splitting these `*_BUILDERS` into separate files? This file is getting a little verbose.

src/parser.ts
```diff
@@ -355,10 +355,20 @@ const EXPR_BUILDER: MarkVisitor<NodeTypes.ExprNode> = {
       return result
     }
 
-    const args: NodeTypes.ExprNode[] = []
+    let args: NodeTypes.ExprNode[] = []
+
     while (p.getMark().name !== 'func_args_end') {
-      args.push(p.process(EXPR_BUILDER))
+      if (argumentShouldBeSelector(namespace, name, args)) {
```
judofyr, line 361: Ah, this code confused me first. The reason this works is because we populate the `args` on the fly and thus it contains all the previous arguments. It's rather strange for the `argumentShouldBeSelector`-function to get access to all of the _previous_ arguments though. What about changing this to:
```suggestion
      if (argumentShouldBeSelector(namespace, name, args.length)) {
```

And then the `argumentShouldBeSelector` actually accepts an `argumentIndex`?

src/parser.ts
```diff
@@ -355,10 +355,20 @@ const EXPR_BUILDER: MarkVisitor<NodeTypes.ExprNode> = {
...
+        // Since the diff/delta functions aren't validated yet we only want to validate the selector
+        // being used. We expect the null valued arg to throw an error at evaluation time.
+        p.process(SELECTOR_BUILDER)
+        args.push(NodeTypes.SelectorNode)
+        break
```
judofyr, line 366: We shouldn't need to `break` here I think 🤔 

```suggestion
```

src/parser.ts
```diff
@@ -355,10 +355,20 @@ const EXPR_BUILDER: MarkVisitor<NodeTypes.ExprNode> = {
...
-    const args: NodeTypes.ExprNode[] = []
+    let args: NodeTypes.ExprNode[] = []
+
     while (p.getMark().name !== 'func_args_end') {
-      args.push(p.process(EXPR_BUILDER))
+      if (argumentShouldBeSelector(namespace, name, args)) {
+        // Since the diff/delta functions aren't validated yet we only want to validate the selector
+        // being used. We expect the null valued arg to throw an error at evaluation time.
+        p.process(SELECTOR_BUILDER)
+        args.push(NodeTypes.SelectorNode)
```
judofyr, line 365: Doesn't this have to be like this (since `SelectorNode` is an interface)?

```suggestion
        args.push({type: 'Selector'})
```

test/parse.test.ts
```diff
@@ -52,6 +52,52 @@ t.test('Error reporting', async (t) => {
...
+      'diff::changedAny({}, {}, (foo))', // BUG: fails because we don't support groups yet
+    ]
+
+    queries.forEach((query) => {
+      try {
+        parse(query)
+      } catch (error: any) {
+        t.same(error.message, 'Invalid selector syntax')
+      }
+    })
```
judofyr, line 85: We prefer native for-of loops for these cases:

```suggestion
    for (const query of queries) {
      try {
        parse(query)
      } catch (error: any) {
        t.same(error.message, 'Invalid selector syntax')
      }
    }
```

judofyr: We can also use [`t.throws`](https://node-tap.org/docs/api/asserts/#tthrowsfn-expectederror-message-extra) here. Something like `t.throws(() => parse(query), 'Invalid selector syntax')`

src/parser.ts
```diff
@@ -662,6 +672,165 @@ const TRAVERSE_BUILDER: MarkVisitor<(rhs: TraversalResult | null) => TraversalRe
...
+  },
+
+  slice(p) {
+    p.process(TRAVERSE_BUILDER)
+    return null
+  },
+
+  group(p) {
+    // This should only throw an error until we add support for groups in selectors.
+    throw new Error('Invalid selector syntax')
```
judofyr, line 688: It's very easy adding support though:

```suggestion
    p.process(SELECTOR_BUILDER)
```

src/parser.ts
```diff
@@ -662,6 +672,165 @@ const TRAVERSE_BUILDER: MarkVisitor<(rhs: TraversalResult | null) => TraversalRe
...
+  dblparent(p) {
+    throw new Error('Invalid selector syntax')
+  },
+
+  traverse(p) {
+    // p.unshift()
+    const _ = p.process(EXPR_BUILDER)
+    p.process(TRAVERSE_BUILDER)
+    return null
+  },
```
judofyr, line 712: Two things here:

1. Right now it only processes the first traversal entry and will leave the parser in a weird state. We need to loop over until we find the `traversal_end`.
2. The base should be parsed using the selector builder to validate it properly.

```suggestion
  traverse(p) {
    p.process(SELECTOR_BUILDER)
    while (p.getMark().name !== 'traversal_end') {
      // TODO: This now starts parsing traversals as regular expressions.
      // We need a special builder for handling just the subset for selectors.
      p.process(TRAVERSE_BUILDER)
    }
    p.shift()
    return null
  },
```

src/parser.ts
```diff
@@ -662,6 +672,165 @@ const TRAVERSE_BUILDER: MarkVisitor<(rhs: TraversalResult | null) => TraversalRe
...
+  this_attr(p) {
+    p.processString()
+    return null
+  },
+
+  attr_access(p) {
+    p.processString()
+    return null
+  },
+
```
judofyr, line 723: We don't need this here. `attr_access` is only valid inside the traversal-context and we only use `SELECTOR_BUILDER` for expressions. I can look into improving the typing here.

```suggestion
```

src/parser.ts
```diff
@@ -662,6 +672,165 @@ const TRAVERSE_BUILDER: MarkVisitor<(rhs: TraversalResult | null) => TraversalRe
...
+const SELECTOR_BUILDER: MarkVisitor<null> = {
+  square_bracket(p) {
+    p.process(TRAVERSE_BUILDER)
+    return null
+  },
+
+  slice(p) {
+    p.process(TRAVERSE_BUILDER)
+    return null
+  },
```
judofyr, line 684: These two are also never called at the moment. Same reason as with `attr_access`.

src/markProcessor.ts
```diff
@@ -82,6 +82,10 @@ export class MarkProcessor {
     this.index += 1
   }
 
+  unshift(): void {
+    this.index -= 1
+  }
+
```
judofyr, line 88: Not used anywhere?

src/parser.ts
```diff
@@ -355,10 +355,20 @@ const EXPR_BUILDER: MarkVisitor<NodeTypes.ExprNode> = {
       return result
     }
 
-    const args: NodeTypes.ExprNode[] = []
+    let args: NodeTypes.ExprNode[] = []
```
judofyr, line 358: No reason to change to `let` here? 🤔

judofyr (changes requested):
Good foundation! I've highlighted some correctness issues which we should tackle before merging this.

As for completing the functionality in itself, I think the best course is to start adding lots of test cases into groq-test-suite to ensure we have parity between Content Lake's implementation and groq-js.

src/parser.ts
codebymatt: Yeah you're absolutely right, I pushed this up at the end of the day and didn't give it a proper review myself - a few messy things slipped through. That's my bad!

src/markProcessor.ts
codebymatt: This is used in `func_call` in the `SELECTOR_BUILDER`. I'll see if there's a way to avoid needing to use it.

test/parse.test.ts
codebymatt: > We prefer native for-of loops for these cases:

Is there a reason why, or is it just our convention?

src/parser.ts
codebymatt: I had these in to catch an error with traversals, but the problem is resolved when I process traversals until the `traversal_end` mark.

test/parse.test.ts
judofyr: (1) It also supports other types of iterations (e.g. async iteration) so it's more consistent around the code base. (2) It makes it easier to add other control flow inside (e.g. `await` or `return`) if needed.

src/parser.ts
```diff
@@ -662,6 +672,153 @@ const TRAVERSE_BUILDER: MarkVisitor<(rhs: TraversalResult | null) => TraversalRe
...
+  },
+
+  tuple(p) {
+    // This should only throw an error until we add support for tuples in selectors.
+    throw new Error('Invalid selector syntax')
+  },
+
+  func_call(p) {
+    p.unshift()
+    const func = p.process(EXPR_BUILDER) as NodeTypes.FuncCallNode
```
judofyr, line 783: Aha! Then we can be more explicit here:

```suggestion
  func_call(p, mark) {
    const func = EXPR_BUILDER.func_call(p, mark) as NodeTypes.FuncCallNode
```

src/parser.ts
judofyr: There's still a break here 🙃

judofyr (commented):
Nice, nice!

src/parser.ts
codebymatt: Ahhh clever!

judofyr (approved):

PR #64, opened 2022-07-21
chore: improve test coverage

codebymatt:
One of the requirements to releasing V1.0.0 of `groq-js` is improved test coverage.

I've added a good few tests to cover functionality that wasn't being tested. A lot of the tests are checking for proper exception handling. Normally I wouldn't test exceptions thoroughly, but I opted to here for the sake of completeness.

See below coverage chart. We're not at 100% test coverage, but getting there would be far more trouble than it's worth.
The vast majority of functions and lines have been tested.

```
-------------------------|---------|----------|---------|---------|-------------------
File                     | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-------------------------|---------|----------|---------|---------|-------------------
All files                |   96.56 |    93.07 |   99.63 |   99.26 |
 src                     |    94.6 |    88.67 |     100 |   99.75 |
  index.ts               |     100 |      100 |     100 |     100 |
  markProcessor.ts       |     100 |      100 |     100 |     100 |
  parser.ts              |   98.96 |    95.93 |     100 |   99.63 | 246
  rawParser.js           |   91.74 |    84.78 |     100 |   99.79 | 306
  traversal.ts           |     100 |      100 |     100 |     100 |
 src/evaluator           |   98.72 |    97.58 |   99.24 |   98.81 |
  constantEvaluate.ts    |   91.66 |       90 |     100 |   91.66 | 50,61
  equality.ts            |     100 |      100 |     100 |     100 |
  evaluate.ts            |    99.5 |    97.88 |     100 |    99.5 | 127
  functions.ts           |   99.59 |    98.98 |   97.14 |   99.58 | 235
  index.ts               |     100 |      100 |     100 |     100 |
  matching.ts            |     100 |      100 |     100 |     100 |
  operators.ts           |   97.64 |    97.77 |     100 |   97.53 | 72,129
  ordering.ts            |     100 |      100 |     100 |     100 |
  pt.ts                  |   96.15 |    95.83 |     100 |     100 | 30
  scope.ts               |     100 |      100 |     100 |     100 |
  scoring.ts             |      95 |    93.54 |     100 |   94.73 | 68,72
 src/values              |   98.41 |    94.44 |     100 |   98.38 |
  Path.ts                |     100 |      100 |     100 |     100 |
  StreamValue.ts         |     100 |      100 |     100 |     100 |
  dateHelpers.ts         |     100 |      100 |     100 |     100 |
  index.ts               |     100 |      100 |     100 |     100 |
  types.ts               |     100 |      100 |     100 |     100 |
  utils.ts               |   96.15 |    91.66 |     100 |   96.07 | 32,131
 tap-snapshots/test      |     100 |      100 |     100 |     100 |
  parse.test.ts.test.cjs |     100 |      100 |     100 |     100 |
-------------------------|---------|----------|---------|---------|-------------------
```


src/rawParser.js
```diff
@@ -765,7 +765,6 @@ function parseString(str, pos) {
             marks.push({name: 'unicode_hex', position: pos + 3})
             pos = str.indexOf('}', pos + 3)
             marks.push({name: 'unicode_hex_end', position: pos})
-            pos++
```
codebymatt, line 768: I think I found a bug in unicode hex parsing, I couldn't get this case (unicode characters delimited with braces) to work without this change. I can discuss this change offline if needs be.

judofyr, line 768: Nice find!

test/testUtils.ts
```diff
@@ -0,0 +1,9 @@
+import t from 'tap'
+
+export async function throwsWithMessage(funcUnderTest: () => {}, expectedMessage: string) {
```
judofyr, line 3: What's the difference of this vs just using `t.throws`?

judofyr, line 3: I _think_ this should take a `t` as a parameter and not use the global `t` instance. When you're using sub-tests `tap` can be confused if you use the wrong `t`.

judofyr (commented):
Nice work! In the future we might want to transition some of the "invalid queries" into groq-test-suite.

test/testUtils.ts
codebymatt: `t.throws` wasn't checking the error message as well. (It might be that I was using it incorrectly.)

codebymatt: It originally took the tap instance as a parameter but I didn't see any use in keeping it.
That's a good point about the sub-tests though, it might mess up reporting the failures. I'll add it back in.

codebymatt: I played around with a few different arguments to the `tap` call, but couldn't get it to match the error message thrown. I may be missing something obvious [in the docs](https://node-tap.org/docs/api/asserts/#tthrowsfn-expectederror-message-extra).

test/testUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export async function throwsWithMessage(
+  t: Tap.Test,
+  funcUnderTest: () => {},
+  expectedMessage: string
+) {
+  try {
+    await funcUnderTest()
+  } catch (error: any) {
+    t.same(error.message, expectedMessage)
```
judofyr, line 11: Hmm. Won't this succeed if the function _doesn't_ throw as well? 🤔

codebymatt: Excellent point. I could call the function twice, once to ensure the exception happens, and again to check the error message, but that feels wasteful.

I'll see if there's a built-in way to test the exact message being thrown, a deeper dive into the docs might be necessary.

judofyr (approved):

PR #66, opened 2022-07-26
chore: document release process and relevant commands

codebymatt:
* Add release steps to the README
* Add instructions regarding the use of the `np` package

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ let tree = parse(input)
 
 let dataset = [
   {_type: 'user', name: 'Michael'},
-  {_type: 'company', name: 'Bluth Company'}
+  {_type: 'company', name: 'Bluth Company'},
 ]
 
 // Evaluate a tree against a dataset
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,17 +50,47 @@ See [API.md](API.md) for the public API.
...
+2. Update the version field in `package.json` and commit it to version control
+3. Create a Git tag with the version number (`git tag vX.Y.Z`)
+4. Push upstream to the `main` branch of the GitHub repo
+5. Run `npm publish`
+6. Check the package was successfully published
+
+We use [this NPM package](https://github.com/sindresorhus/np) to handle steps 2 through 5.
+You need to install it on your system with:
+
+```bash
```
sgulseth, line 77: How about we add `np` to devDependencies and then use `npx np ...`? That way we use the same version across users and publishes

sgulseth (changes requested):

README.md
```diff
@@ -50,17 +50,44 @@ See [API.md](API.md) for the public API.
...
+6. Check the package was successfully published
+
+We use [this NPM package](https://github.com/sindresorhus/np) to handle steps 2 through 5.
+It's included as a dev dependency, you can install it with your package manager of choice by running the install command (see [Installation](#Installation)).
+It will be installed by default when you set this repo up for the first time.
+
+You can interactively update the version by running the following from the project root. Note that you'll need NPM publish permissions. We need to specify we don't want to create a GitHub release draft.
+
+```bash
+np --no-release-draft
```
sgulseth, line 81: We should also prefix L81 & L87 with `npx`. By default the system wont look up binaries in `node_moduels/.bin`, but `npx` will

sgulseth (changes requested):

README.md
codebymatt: Oh interesting! Yeah, that's a sensible thing to do

README.md
```diff
@@ -50,17 +50,44 @@ See [API.md](API.md) for the public API.
...
+## Releasing a new version
+
+The following steps are required to release a new version. You should only release from the `main` branch.
+
+1. Update the CHANGELOG inline with the existing format (remember [SemVer](https://semver.org)!)
+2. Update the version field in `package.json` and commit it to version control
+3. Create a Git tag with the version number (`git tag vX.Y.Z`)
+4. Push upstream to the `main` branch of the GitHub repo
+5. Run `npm publish`
+6. Check the package was successfully published
```
sgulseth, line 72: This can be shortened a bit with `np`

1. Update the CHANGELOG inline with the existing format (remember [SemVer](https://semver.org)!)
2. Run `npx np`
  	`np` will run the tests,  update package.json, create a git tag, push it to github and publish the package on npm

sgulseth (changes requested):
Sorry for the cumbersome review here, missed a few things in the first review.

Isn't the point to use `np` for publishing and not just verifying?

codebymatt:
> Isn't the point to use `np` for publishing and not just verifying?

For sure, the command on L81 will do the publishing.

I added L87 since being able to do dry-runs is generally helpful. I can reword this section if it's unclear though.

README.md
sgulseth: But if you use `np` you shouldn't create a git tag since `np` will do it, same goes for updating package.json.

We can rather document what `np` is doing?

codebymatt: If someone wants to understand the explicit steps to how we release `groq-js` they won't be able to figure that out by checking the README, they'll need to go read what `np` does.

That seems like uneccessary redirection when we could lay out the required steps, and then suggest they use the tool to make life easier.

I'll reword this to point readers to `np`, but with a brief explanation of the underlying steps.

codebymatt: > We can rather document what np is doing?

Ahhhh that's what you suggested anyways!

judofyr (approved):
LGTM!

sgulseth (approved):
Nice!

PR #67, opened 2022-07-28
release: v1.0.0-rc

codebymatt:
ref: sc-20405

This is relatively straightforward, we need to update the CHANGELOG before cutting a release candidate.

Feedback welcomed on wording/presentation!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## v1.0.0-rc - 2022-07-28
+
+New features:
+
+- Implement `array::unique`
+- Implement aggregation functions on the `math` namespace:
+  - `min()`
+  - `max()`
+  - `sum()`
+  - `avg()`
+- Implement functions on the `string` namespace:
+  - `startsWith()`
+  - `split()` (experimental)
+- Implement functions on the `array` namespace:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+## v1.0.0-rc - 2022-07-28
+
+New features:
+
+- Implement `array::unique`
```
atombender, line 7:
```suggestion
- Implement `array::unique()`
```

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+  - `min()`
+  - `max()`
+  - `sum()`
+  - `avg()`
+- Implement functions on the `string` namespace:
+  - `startsWith()`
+  - `split()` (experimental)
+- Implement functions on the `array` namespace:
+  - `join()` (experimental)
+  - `compact()` (experimental)
```
atombender, line 18: Technically all of the functions here are experimental. We don't make a distinction in `groq-js` at the moment, though, only in the API.

Either we should drop "experimental" here, or we should invent some kind of internal feature flag for it, and I suggest we don't do the latter at this point.

atombender (changes requested):

PR #68, opened 2022-08-04
Fix build error related to selectors

judofyr:
The build failed locally for me due to this missing evaluator.

Also added `npm run build` to the CI to ensure that we don't break the build.


codebymatt (approved):
Good catch!

PR #69, opened 2022-08-04
Switch over to new GROQ versioning scheme

judofyr:
Move the current GROQ implementation into 1.ts to clarify it implements GROQ-1. However, we still only support importing directly from `groq-js`.

When GROQ-2 comes along we can figure out exactly how we want to expose both implementations in a single package.


README.md
```diff
@@ -3,7 +3,7 @@
 GROQ-JS is a JavaScript implementation of [GROQ](https://www.sanity.io/docs/data-store/how-queries-work) which follows the official specification.
 
 ```javascript
-import {parse, evaluate} from 'groq-js/v2'
+import {parse, evaluate} from 'groq-js/1'
```
codebymatt, line 6: Should we add a line or two explaining the need to import based on GROQ versions? I know we don't have a V2 yet, but we can explain how it works.

codebymatt (approved):

codebymatt:
> we might as well make the "default version" always refer to the last one.

Asking because I'm not famailair with this side of package management: where in the config points the default version to the latest one?

judofyr:
> > we might as well make the "default version" always refer to the last one.
> 
> Asking because I'm not famailair with this side of package management: where in the config points the default version to the latest one?

https://github.com/sanity-io/groq-js/blob/d911b6d5072fe7ff04e225f00f842b92beb44a57/package.json#L9. We'd have to change this to refer to `1`.

codebymatt:
> We'd have to change this to refer to `1`.

So changing this to refer to `1` would lock the version to GROQ-1, but leaving it as `.` ensures the latest GROQ version is used?

codebymatt (approved):

sgulseth (approved):

PR #70, opened 2022-08-08
fix: ensure 'default' export path is specified last 

codebymatt:
I was testing the new functionality inside another project (GROQ Arcade). The project wouldn't build and was erroring at runtime. It turns out Webpack is _very_ fussy about certain things, like whether the 'default' for exports is last in the list or not.

Changing the order fixes the issue. As far as I can tell this shouldn't negatively impact other builds/projects, so I think we should include the change in `main` and a new release candidate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,12 +8,12 @@
   "exports": {
     ".": {
       "import": {
-        "default": "./dist/1.esm.mjs",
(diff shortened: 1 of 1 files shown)


judofyr (approved):
I have no clue how this would help Webpack as the order of fields in an object _really_ shouldn't matter 🤷

PR #72, opened 2022-08-16
fix: avoid "instanceof ConstantEvaluateError"

judofyr:
I'm not entirely sure of the problem here, but in the released version of groq-js 1.0.0 in Node.js 18 I'm seeing a strange error where it appears that "err instanceof ConstantEvaluateError" returns false. This is a bit hard to test for since it seems to be caused by the way Rollup builds the file.


judofyr:
Oops. My first fix didn't work. Pushed the full fix now.

codebymatt (approved):

PR #78, opened 2022-11-15
chore: add failing repro test case

stipsan:
Running this query on Content Lake works fine:
```groq
count(array::unique(
  *[_type == 'page']{"_id": select(
    _id in path("drafts.**") => _id,
    "drafts." + _id
  )}._id
))
```

But creates a parse error in `groq-js`, attached a repro test case in this PR.

### DIFF
--- a/test/parse.test.ts
+++ b/test/parse.test.ts
@@ -37,6 +37,18 @@ t.test('Basic parsing', async (t) => {
     const tree = parse(query)
     t.matchSnapshot(tree)
   })
+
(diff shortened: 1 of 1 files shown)


j33ty:
The problem with JS parser is that it does not consume any whitespace in front of the function argument, for namespaced functions. So, a simple query that would also fail before this PR: `pt::text( *.text)` while `pt::text(*.text)` will pass.

Apologies for taking liberty to push a fix on the same branch.

PR #79, opened 2022-12-28
fix: use pkg-utils to generate JS and typings

stipsan:
Also solves issues with some types not being exported. Should probably be a major new version.

Testable using `npm i groq-js@beta`

Build output diff: https://app.renovatebot.com/package-diff?name=groq-js&from=1.1.4&to=1.2.1-beta.1


package.json
```diff
@@ -10,30 +10,52 @@
...
+        "import": "./dist/index.cjs.mjs",
+        "require": "./dist/index.js"
       },
-      "require": {
-        "types": "./dist/1.umd.d.ts",
-        "default": "./dist/1.umd.cjs"
+      "import": "./dist/index.mjs",
+      "default": "./dist/index.js"
+    },
+    "./1": {
```
stipsan, line 26: This lets consumers pin their `groq-js` to a specific GROQ version. While the main `.` export can just always point to the latest spec that's implemented.
Also worth noting if an app ends up doing both `import 'groq-js'` and `import 'groq-js/1'` our `@sanity/pkg-utils` tooling is smart enough to dedupe code between the export points and you don't end up doubling the memory usage and amount of code that's loaded up.

In fact, `dist/index.mjs` currently looks like this, since both exports use the same exact entrypoint:
```js
export { evaluate, parse } from './1.mjs';
```

package.json
```diff
@@ -10,30 +10,52 @@
...
   "exports": {
     ".": {
-      "import": {
-        "types": "./dist/1.esm.d.ts",
-        "default": "./dist/1.esm.mjs"
+      "types": "./dist/index.d.ts",
+      "source": "./src/1.ts",
+      "require": "./dist/index.js",
+      "node": {
+        "import": "./dist/index.cjs.mjs",
```
stipsan, line 20: This is a generated file by `@sanity/pkg-utils` that looks like this:
```js
import cjs from './index.js';

export const evaluate = cjs.evaluate;
export const parse = cjs.parse;

```

This Node.js specific problem this technique solves is known as the [Dual Package Hazard](https://nodejs.org/api/packages.html#dual-package-hazard) and is known to cause `instanceof` checks and similar to break if both the ESM and CJS version of a package is loaded up and used interchangeably in an app. Usually due to long chains of dependencies of dependencies that go in and out of the CJS and ESM runtime.

package.json
```diff
@@ -10,30 +10,52 @@
...
   },
-  "main": "./dist/1.umd.cjs",
-  "types": "./dist/1.umd.d.ts",
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
+  "source": "./src/1.ts",
+  "types": "./dist/index.d.ts",
+  "typesVersions": {
+    "*": {
+      "1": [
```
stipsan, line 45: This is necessary for TypeScript to find the type definitions for `import {parse, evaluate} from 'groq/1'` imports. TS doesn't yet support `pkg.exports.types`. TypeScript's new `Node16` mode supports `pkg.exports.types` but it doesn't count as the way it requires you to write ESM import paths isn't compatible with other tooling like ESLint, babel, rollup and other bundlers.

package.json
```diff
@@ -47,24 +69,29 @@
       "esbuild-register"
     ]
   },
+  "browserslist": [
+    "> 0.2% and supports es6-module and supports es6-module-dynamic-import and not dead and not IE 11",
```
stipsan: We use the same browserlist target for the rest of our ecosystem.

judofyr (approved):
I don't fully understand the full impact of all of these choices, but as long as tests pass and the package builds, I'm happy!

PR #94, opened 2023-02-13
Add types annotations

OFRBG:


OFRBG:
Hello! Sorry about opening this PR here. I was using codespaces and it targeted upstream instead of origin by default.

PR #162, opened 2024-01-17
chore: upgrade prettier

judofyr:


judofyr:
This also re-runs prettier. Not sure the tests are currently broken on `main`.

PR #175, opened 2024-02-21
feat: can extract attribute key from Group

sgulseth:
Need to make sure we align with https://sanity-io.github.io/GROQ/draft/#DetermineName()


judofyr:
Can we add a test for this behavior? Is there a GROQ query which groq-js doesn't parse correctly?

judofyr:
Oh, interesting. Seems like a bug in GROQ spec as well? `Group` isn't mentioned there.

sgulseth:
mhm. I'll add group and the other missing types

judofyr (approved):

PR #176, opened 2024-02-26
feat: add type evaluator

sgulseth:
Adds:
	* Set of NodeType that represent a type structure
	* New evaluator that returns types for a given query and schema represented as a list of Documents and TypeDeclaration. 

What to review:
	* Do we traverse and handle the different Nodes correctly?
	* Do the tests cover and represent expected functionality? 
	* Can we improve readability? 

Type evaluation expects a schema generated by https://github.com/sanity-io/sanity/pull/5814


tzhelyazkova (commented):
just FYI: I'm going to work on more tests for this logic (and base it on top of https://github.com/sanity-io/groq-js/pull/178).

judofyr:
Some thoughts after looking through this. (With some caveat that I might not have the full context here)

## Public API

### Proposal: `evaluateQueryType` to work on AST instead of a string.

I don't see any reason to couple parsing into the evaluator. This also came up in #179 since there I explicitly work on the AST and not on a string.

### In general: Maybe we should add `{type: string}` as the public API and remove `UnknownType`?

There's a nice way of making the API forwards-compatible: If we say that `TypeNode` also includes `{type: string}` as a potential type and that all consumers should treat this as "unknown". This means that we can add support for new TypeNode in minor versions without worrying about anything crashing. If we're doing this, then we should probably also remove `{type: "unknown"}` as an exported type since users shouldn't actually depend on it.

### `optional`

It's not clear to me what this actually means today. Is this just a shortcut for a `union` with a `null`? Or does it also imply something about the presence of a value in an object?

Proposal: Remove (and see next point with the replacement).

### `object`

Note that GROQ currently refers to the collection of key/value as "attributes" and not "fields". I know that Studio uses the term "fields", but since we're in GROQ-land I believe we should continue using "attributes". A few more comments:

- I don't see why `fields` should be an array. The reason it's an array in the `ObjectNode` is because there the other matters, but in this context we don't care about it.
- We're currently lacking a way of specifying "a generic object (with unknown attributes)". This is a bit unfortunate since there are many cases where we can actually know that. E.g. `foo{...bar}` is guaranteed to be either `null` or an `object` (regardless of the values of )
- We're also lacking a way of specifying "an object with these attributes, but also something unknown". I think this would be useful.

Proposal: Replace with the following:

```ts
type ObjectTypeNode = {
  attributes: Record<string, ObjectAttribute>
  fallback?: TypeNode
}

type ObjectAttribute = {
  type: TypeNode
  optional: boolean
}
```

### `concatenate`

This seems like a very narrow use case. It's also not obvious to me if it's supposed to handle all types of concatenation (arrays? objects?).

Proposal: Remove.

### `never`

This doesn't actually seem to be generated. I'm not sure if it makes so much sense for GROQ since we typically always end up with a value (although possibly `null`).

### `parameter`

This seems a bit halfway done now. It currently doesn't have any type attached to it.

Proposal: Remove. Return `{type: "unknown"}` for now.

### `reference`

This one is … wonky. It seems to cover to different cases:

1. Handling the concept of "named types" which are embedded values.
2. Handling of document references.

I don't understand why these are coupled together.

It also seems to be unable to reason about that references are also objects (e.g. `foo._ref` should be a string).

Proposal:

```ts
/**
 * These are the options passed into `evaluateQueryType` and replaces "schema":
 */
type Options = {
  dataset: TypeNode
  aliases: Record<string, TypeNode>
}
/**
 * Alias type refers to one of the named aliases (given to the options).
 */
type AliasType = {
  type: 'alias'
  name: string
}

/**
 * Represents a value that can be dereferenced.
 */
type DereferenceType = {
  type: 'reference'
  of: TypeNode
  dereferenceType: TypeNode
}
```

Then we can generate the following for a Sanity schema:

```ts
const options: {
  aliases: {
    ref: {
      type: 'ObjectType'
      attributes: {
        _ref: {
          type: {type: 'string'}
          optional: false
        }
        _weak: {
          type: {type: 'boolean'}
          optional: true
        }
      }
    }
    'document.author': {
      type: 'ObjectType'
      attributes: {
        //
      }
    }
    'document.book': {
      type: 'ObjectType'
      attributes: {
        author: {
          type: {
            type: 'reference'
            of: {type: 'alias'; name: 'ref'}
            dereferenceType: {type: 'alias'; name: 'document.author'}
          }
          optional: true
        }
      }
    }
  }
}
```

sgulseth:
### `evaluateQueryType to work on AST instead of a string.`
:+1: agree. would also align the API with evaluate.

### In general: Maybe we should add {type: string} as the public API and remove UnknownType?

Want to push back a bit on this. It makes it harder to catch and handle types correctly parsing the type nodes, as any type name would be valid. There should be a known case of types, which new types do you think would be added in the future?


### `optional`

remove and add as a property on objectAttribute. Agree. No strong opinions, was a bit back and forth with this previously, and felt the optional-type was easier to reason about then, but now that the type nodes are more fleshed out I don't see the benefit.

### `concatenate`

remove. :+1: - no strong opinions.

### `never`

remove :+1: can just be null/not exist

### `parameter`

can be unknown for now :+1:

### `object`

Attributes is a better name than fields. :+1: - agree.
Moving optional to objectAttribute is a good idea. :+1: - agree.

> I don't see why fields should be an array. The reason it's an array in the ObjectNode is because there the other matters, but in this context we don't care about it.

hmm. Doesn't the order matter if it's a projectino with attribute access and spreads? :thinking: I agree and can't think of anything right now, but I'm pretty sure I had a reason for it. :sweat_smile: I can refactor it out, and if we find a reason for it later we can add it back in.

> We're currently lacking a way of specifying "a generic object (with unknown attributes)". This is a bit unfortunate since there are many cases where we can actually know that. E.g. foo{...bar} is guaranteed to be either null or an object (regardless of the values of )
&
> We're also lacking a way of specifying "an object with these attributes, but also something unknown". I think this would be useful.

Can't this just then be an `unknown` for now?

### `reference`

> I don't understand why these are coupled together.

The reason is to make it easier to extract them from the sanity schema into a json representation. As in the schema we just know that it's a reference to either a document or a "type". 

However, I agree that it's confusing to couple them together, and it would be good to separate them.
1. We don't actually need to have a concept of "type"/alias in the schema at all, as we could just embed them directly into the type nodes. Dowside here is that the generated schema would be more verbose, and so would the generated types as we don't reuse those types across socuments.
2. I think that if we are to separate them into different nodes, we should also separate the referencing node type. You don't have to dereference an "alias", in content lake it's always just stored as any other type(Usually object or array of objects, but I've also seen ie. strings). 

---

Since you are replacing Schema with Options, would that mean that schema extract would then rather extract the set of options? Meaning the response from schema extract would be an object like:

```{
  "dataset": TypeNode
  aliases: Record<string, TypeNode>
}
```

Or do we parse the returned schema and derive the Options from that? 

Also would `dataset` always be a union of documents?

---

Bike shedding a bit: Alias feels like a new concept, could it just be "types"?

judofyr:
> Want to push back a bit on this. It makes it harder to catch and handle types correctly parsing the type nodes, as any type name would be valid. There should be a known case of types, which new types do you think would be added in the future?

`dateTime` and `geo` are maybe the most obvious ones. These are serialized to JSON, but the JSON itself is limited. Also, something like `concatenation` could be a 

However, on second thought, maybe we should then instead structure it around the JSON type. We currently have these "fundamental" serialization types: `number`, `boolean`, `string`, `array`, `object`. Everything in the ends up as these. So maybe the right way of adding support for these would be to return `{type: "string"; kind: "dateTime"}` in the future. From a serialization point of view it will show up as a string. From an evaluation point of view it's very different of course.

So yes, let's scrap this proposal. Most consumers of the types cares about the JSON representation, not the internal one (since they retrieve it from a serialized API).

> hmm. Doesn't the order matter if it's a projectino with attribute access and spreads? 🤔 I agree and can't think of anything right now, but I'm pretty sure I had a reason for it. 😅 I can refactor it out, and if we find a reason for it later we can add it back in.

From a type perspective I don't think the order matters. `TypeNode` is actually closer to a `Value`. E.g. the regular evaluate function executions the object in order, but then returns an object with a "flat" list of properties. And type evaluation should then do the same thing.

> Can't this just then be an unknown for now?

This actually came up in #179. In one of the case of `+` we returned `{type: "number"}` in the fallback-case, but in reality it will only be `number`, `array`, `string`, `object` (since these support `+`) or `null` (due to type mismatch). However, since there was no way of representing "a generic object" then we have just return `unknown` here.

However, this also touches on another point which is a bit unclear to me: If the type is `object`, do we assume that it's "closed" or "open" with respect to the attributes? E.g. if you access an attribute which isn't part of the defined list, should that return `unknown` (because we don't know anything about the others) or `null` (because we know that we have the complete list).

I think it would be _very_ valuable if for the cases where we _fully_ know the fields (e.g. `{foo, bar}`) we are able to reason about that. I can also see value in maybe the Sanity Studio wanting to define all of its objects as "extendable".

> We don't actually need to have a concept of "type"/alias in the schema at all, as we could just embed them directly into the type nodes. Dowside here is that the generated schema would be more verbose, and so would the generated types as we don't reuse those types across socuments.

They can be recursive though. E.g. "author" can have references to other "authors"? Then we can't embed them directly.

> Bike shedding a bit: Alias feels like a new concept, could it just be "types"?

Yea, maybe they are just named types?

judofyr:
I've been thinking more about the problem of "open/closed objects" and I have two proposal.

S(Here I'm also assuming we have `{type: "alias", name: "…"}` for referring to a named type and `{type: "dereference", to: …}` for saying "when dereferencing this value it returns this type", but that is not part of this proposal.)

## Merge operator

```ts
/**
 * The merge type allows you to represent a type combined of multiple other
 * types. Most notably, you can merge object types to represent the result of
 * merging these types. The order is significant.
 */
type MergeTypeNode = {
  type: 'merge'
  of: TypeNode[]
}
  // Attaching `_key: string` to another object:
  const withKey: MergeTypeNode = {
    type: "merge",
    of: [
      {
        type: 'object',
        attributes: {_key: {type: 'objectAttribute', value: {type: 'string'}}},
      },
      {type: 'alias', name: 'author'}
    ]
  }
 
  // Representing a reference:
  const ref: MergeTypeNode = {
    type: "merge",
    of: [
      // Both an object
      {
        type: 'object',
        attributes: {_ref: {type: 'objectAttribute', value: {type: 'string'}}},
      },
      // … and something which dereferences to another type
      {type: "dereference", to: {type: 'alias', name: 'author'}}
    ]
  }
 
  // Representing an object where we don't know all of the attributes:
  const open: MergeTypeNode = {
    type: "merge",
    of: [
      // Both an object with some known attributes,
      {
        type: 'object',
        attributes: {a: {type: 'objectAttribute', value: {type: 'string'}}},
      },
      // … but also a generic object.
      {type: "object"}
    ]
  }
```

(Note that this also depends on `attributes` being optional on the object-type and then it represents a "generic object with unknown keys")

## Fallback on `object`

```ts
/**
 * Represents an object, with the given set of attributes. An object can also
 * have a "fallback" which represents the "rest" of the object excluding the
 * specified attributes.
 * 
 * If `fallback` is not present, then it represents that we know all the
 * attributes in the object, and accessing any other attribute will return
 * `null`.
 */
type ObjectTypeNode = {
  type: 'object'
  attributes?: Record<string, ObjectAttribute>
  fallback?: TypeNode
}

// This can be used to add extra attributes to an existing object type: 
const withKey: ObjectTypeNode = {
  type: 'object',
  attributes: {_key: {type: 'objectAttribute', value: {type: 'string'}}},
  fallback: {type: 'alias', name: 'slug'},
}

// A reference is both an object, but falls back to a "dereference" type.
const ref: ObjectTypeNode = {
  type: 'object',
  attributes: {_ref: {type: 'objectAttribute', value: {type: 'string'}}},
  fallback: {type: 'dereference', to: {type: 'alias', name: 'author'}},
}

// An "open" object falls back to `unknown` such that any AccessAttribute on it also becomes "unknown".
const open: ObjectTypeNode = {
  type: 'object',
  attributes: {a: {type: 'objectAttribute', value: {type: 'string'}}},
  fallback: {type: 'unknown'},
}
```

sgulseth:
I like the utility, and extendability, that proposal 1, a MergeObject give, however if we are to give with it I would prefer that the interface would be something like:
```ts
type MergeTypeNode = {
  type: 'merge'
  of: (ObjectTypeNode | DereferenceTypeNode | UnknownTypeNode | AliasTypeNode)[]
}
```

This to reduce the scope when implementing support for it, and it's not 100% clear what would happen if you extended Object and String for example, I guess it's either `null` or just ignored? Could of course also say that merging any other type than the ones defined returns `null` or ignores it.

Bike shedding: `CombineTypeNode`

---

That said, I have nothing against proposal 2 either! Though it feels a bit ambiguous that "fallback" acts as both a "shadow operator"(For handling dereferences) and a "rest operator"(For merging a object onto itself). Is there any benefits of combining alias and dereference? Could it be something like this:

```ts

interface AliasTypeNode {
  type: 'alias'
  name: string
}
type ObjectTypeNode = {
  type: 'object'
  attributes: Record<string, ObjectAttribute>
  rest?: ObjectTypeNode | UnknownTypeNode | AliasTypeNode
  ref?: ReferenceTypeNode
}

// This can be used to add extra attributes to an existing object type:
const withKey: ObjectTypeNode = {
  type: 'object',
  attributes: {_key: {type: 'objectAttribute', value: {type: 'string'}}},
  rest: {type: 'alias', name: 'slug'},
}

// When a ref-property is present it can be Dereferenced by the -> operator, if ref is undefined it returns `null`
const ref: ObjectTypeNode = {
  type: 'object',
  attributes: {_ref: {type: 'objectAttribute', value: {type: 'string'}}},
  ref: {type: 'reference', name: 'author'},
}

// An "open" object falls back to `unknown` such that any AccessAttribute on it also becomes "unknown".
const open: ObjectTypeNode = {
  type: 'object',
  attributes: {a: {type: 'objectAttribute', value: {type: 'string'}}},
  rest: {type: 'unknown'},
}

```

bike shedding: wdyt about `inline` instead of `alias`? As what it's doing is to inline a type declaration in its place?

```ts
interface InlineTypeNode {
  type: 'inline'
  name: string // this references a type declared by TypeDecleration
}
```

judofyr:
+1 on `inline`, and +1 on `rest`.

Good point about references. Currently all references are objects and I don't see any reason why we don't always want to do that. How about removing the `ReferenceTypeNode` and just have the `dereferencesTo`?

```ts
type ObjectTypeNode = {
  type: 'object'
  attributes: Record<string, ObjectAttribute>
  rest?: ObjectTypeNode | UnknownTypeNode | InlineTypeNode
  dereferencesTo?: TypeNode
}

type InlineTypeNode = {
  type: 'inline'
  name: string // this references a type declared by TypeDecleration
}
```

I'm still leaning towards extending `object` just because it means one less type to handle and I still struggle a bit convincing myself of the exact semantics of `merge`.

sgulseth:
Agree on extending object, feels less of a footgun. 

About deReferencesTo, having it as a typnode/ObjectNode would make the schema quite verbose? And would also be slow to build? DeReferencesTo could just be a string though? 🤔

judofyr (approved):

PR #178, opened 2024-02-26
Remove document as a typenode and improve logging

sgulseth:
This removes Document as a TypeNode, keeping the api surface smaller and more actual type oriented


judofyr (approved):
Looks like a good improvement! Let's merge this into #176 and then continue working there instead.

PR #181, opened 2024-02-28
resolve: remove concatenation type

sgulseth:


judofyr (approved):
Changes to the public API looks good to me! I'm focusing mostly on this here and then we can tackle fixing the tests/implementation later.

PR #182, opened 2024-02-29
fix fine grained tests

sgulseth:


test/evaluateQueryNode.test.ts
```diff
@@ -178,14 +180,15 @@ const objects1InArrays = inArray(objects1)
...
-  [primitives, objects1, arrays, objects1InArrays],
+  [primitives, primitivesWithUnknown, objects0, arrays, objects0InArrays],
+  [primitives, primitivesWithUnknown, objects1, arrays, objects1InArrays],
 ]
 
 const SCHEMA: [] = []
 
 const ALL_CATEGORIES = [
   Category.PRIMITIVES,
+  Category.PRIMITIVES_WITH_UNKNOWN,
```
judofyr, line 191: Hmm. This means that when you use `ALL_CATEGORIES` then you actually get all of the primitives twice. I think we either need to:

1. Only have `PRIMITIVES_WITH_UNKNOWN`
2. Or have `PRIMITIVES` and `UNKNOWN`

judofyr (changes requested):
Fixes look good! Left one comment about the `ALL_CATEGORIES` where I think it now generates duplicate variants. If you resolve this then we're good to merge this + #179.

PR #187, opened 2024-03-07
feat: access optional attributes should return null

sgulseth:


test/typeEvaluate.test.ts
```diff
@@ -899,9 +899,9 @@ t.test('deref with projection union', (t) => {
         },
         authorOrGhostName: {
           type: 'objectAttribute',
-          value: {
+          value: nullUnion({
             type: 'string',
-          },
+          }),
```
judofyr, line 904: Studio will never create a schema like this, no?

judofyr (commented):
The logic in `handleAccessAttributeNode` looks sensible, but:

- There doesn't appear to be a test for it? The changes in `typeEvaluate.test.ts` seems to be about a field being a null-union, not optional? There's no test for `optional: true` here as far as I can see.
- I don't see why the changes in `handleOpCall` are needed.

test/typeEvaluate.test.ts
sgulseth: Good point!

sgulseth:
The change is needed because given the op `lhs < 5`, `lhs` could now be an union of `number | null` so we're checking both the potential conditions compared to `5`

test/typeEvaluate.test.ts
sgulseth: Ah, this is not the studio schema, but the projected values. Given that `authorOrGhostName` can be optional, when you are projecting it will be a union of string and null. There shouldn't be any null nodes in the schema

judofyr: Aaah, I see now!

judofyr (approved):

PR #188, opened 2024-03-07
feat: add support for slice in type evaluator

sgulseth:


src/typeEvaluator/typeEvaluate.ts
```diff
@@ -62,14 +63,15 @@ const $warn = debug('typeEvaluator:evaluate::warn')
...
+  $debug('evaluateQueryType.ast %O', ast)
   const parsed = walk({
     node: ast,
     scope: new Scope([], undefined, new Context(schema)),
   })
 
   $trace('evaluateQueryType.parsed %O', parsed)
   const optimized = optimizeUnions(parsed)
-  $trace('evaluateQueryType.optimized %O', optimized)
+  $debug('evaluateQueryType.optimized %O', optimized)
```
tzhelyazkova, line 74: Why debug instead of trace?

tzhelyazkova (approved):

src/typeEvaluator/typeEvaluate.ts
sgulseth: So that you can get some minimal debug-information by turning on debug, you get the ast passed in and the resulting typenode

PR #192, opened 2024-03-11
feat: add tsdoc to describe the type nodes

sgulseth:


src/typeEvaluator/types.ts
```diff
@@ -1,67 +1,121 @@
+/**
+ * Represents a document structure with a fixed type 'document', a name, and a collection of attributes.
+ * Each attribute is defined by a key-value pair where the key is a string and the value is an ObjectAttribute.
```
tzhelyazkova, line 3: shouldn't this sentence about the attribute be right above the `attributes` key?

tzhelyazkova (approved):

src/typeEvaluator/types.ts
sgulseth: Agree!

PR #202, opened 2024-03-26
feat(type-evaluation): add support for Pos and Neg nodes

sgulseth:


j33ty (approved):
LGTM

j33ty:
Slightly curious what inspired this change?

sgulseth:
It's in the back list of operators that we should support, just haven't been prioritizing until now. Had 20 minutes, so chose a low hanging fruits in terms of expressions to implement support for 😄

PR #212, opened 2024-04-05
chore(typeEvaluation): use mapConcrete when handling opCalls

sgulseth:
* Removes `mapUnion`
* Handle cases where you could be comparing `inline` types


judofyr (approved):
Change looks sensible. Any specific bug this was supposed to fix, or just cleaning up?

sgulseth:
Ah! Should add some tests for this as well. 
It fixes a bug where if you try to compare an `inline` type with a value it would always return false since it's comparing values. I'll add tests 😄

sgulseth:
This is mainly for cleaning up. I have a separate branch that builds on this to map concrete values when comparing opcalls

PR #222, opened 2024-04-18
fix(types): use the strictest tsconfig preset

stipsan:
Uses the new `@sanity/pkg-utils/tsconfig/strictest.json` tsconfig preset, and refactors the codebase to align with the stricter rules.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,5 +13,6 @@ module.exports = {
     'no-return-await': 'off',
     'simple-import-sort/imports': 'warn',
     'simple-import-sort/exports': 'warn',
+    'dot-notation': 'off',
   },
 }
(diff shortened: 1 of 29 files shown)


.eslintrc.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
     'no-return-await': 'off',
     'simple-import-sort/imports': 'warn',
     'simple-import-sort/exports': 'warn',
+    'dot-notation': 'off',
```
sgulseth, line 16: Why turn off dot notation? What value does it provide? 🤔

stipsan: The value it provides is ESLint and TypeScript agreeing on what's valid syntax:

https://github.com/sanity-io/groq-js/assets/81981/18195ee0-9c63-4b3a-a091-c625431af679

PR #225, opened 2024-04-24
fix(typeEvaluator): Support math functions

sgulseth:


src/typeEvaluator/functions.ts
```diff
@@ -73,6 +73,29 @@ export function handleFuncCallNode(node: FuncCallNode, scope: Scope): TypeNode {
...
+      // use mapConcrete to get concrete resolved value, it will also handle cases where the value is a union
+      return mapConcrete(values, scope, (node) => {
+        // Aggregate functions can only be applied to arrays
+        if (node.type === 'array') {
+          // Resolve the concrete type of the array elements
+          return mapConcrete(node.of, scope, (node) => {
+            // Math functions can only be applied to numbers
+            if (node.type === 'number') {
+              return node
+            }
```
judofyr, line 90: This won't be correct when `value` is set and the function is `sum` or `avg`, no?

judofyr (commented):
I think I found one problem with the implementation, but otherwise it looks good.

Can we also add tests like this for these?

```diff
diff --git i/test/typeEvaluateCompare.test.ts w/test/typeEvaluateCompare.test.ts
index d6f25e8..0c5b39b 100644
--- i/test/typeEvaluateCompare.test.ts
+++ w/test/typeEvaluateCompare.test.ts
@@ -2,6 +2,7 @@
 import t from 'tap'
 
 import {evaluate} from '../src/evaluator'
+import {namespaces} from '../src/evaluator/functions'
 import {operators} from '../src/evaluator/operators'
 import type {ExprNode, OpCall} from '../src/nodeTypes'
 import type {TypeNode} from '../src/typeEvaluator'
@@ -415,3 +416,16 @@ for (const op of ops) {
     })
   })
 }
+
+t.test('math::max', async (t) => {
+  subtestUnary({
+    t,
+    build: (param) => ({
+      type: 'FuncCall',
+      name: 'max',
+      namespace: 'math',
+      args: [param],
+      func: namespaces.math.max,
+    }),
+  })
+})
```

src/typeEvaluator/functions.ts
sgulseth: good catch, yess

sgulseth:
Nice! I also caught a few more bugs because of some differences between avg and sum :D

judofyr (approved):

PR #226, opened 2024-04-30
feat(typeEvaluator): handle not nodes when resolving conditions

sgulseth:


src/typeEvaluator/typeEvaluate.ts
```diff
@@ -821,9 +821,15 @@ function resolveCondition(expr: ExprNode, scope: Scope): boolean | undefined {
...
+    case 'AccessAttribute':
+    case 'AccessElement':
     case 'Value': {
       const value = walk({node: expr, scope})
-      return value.type === 'boolean' && value.value !== false
+      if (value.type === 'boolean') {
+        return value.value
+      }
+
+      return false
```
judofyr, line 832: I'm not sure if this makes sense for `AccessAttribute`. E.g. if evaluating `AccessAttribute` returns `unknown` then we want to return `undefined`.

Also: Can't `resolveCondition` just be like this? 🤔 

```ts
const value = walk({node: expr, scope})
if (value.type === 'boolean') return value.value
if (value.type === 'null') return false
return undefined
```

And then we depend on the regular `walk` function to handle the different operators?

sgulseth: Agree on returning undefined on unknown!

---

We would need to refactor it a bit as handleOpCall depends on resolveCondition 🤔

judofyr: Hm. This doesn't seem to have been addressed in the latest re-review? 🤔

sgulseth: Updated it to default to `undefined` but also return false on array, object, null

judofyr (approved):

PR #228, opened 2024-05-02
Hoist duplicate type nodes and fix too aggressive flattening 

sgulseth:
We were flattening a bit too much, we should only flatten if it's one union inside another union.

Should produce: `(a,b,c,d) | (a,b,e,f) = (a, b), ((c,d) | (e,f))`

### DIFF
--- a/src/typeEvaluator/optimizations.ts
+++ b/src/typeEvaluator/optimizations.ts
@@ -62,9 +62,66 @@ export function removeDuplicateTypeNodes(typeNodes: TypeNode[]): TypeNode[] {
   return newTypeNodes
 }
 
+export function hoistDuplicateTypeNodes(src: TypeNode[]): TypeNode[] {
+  const times = new Map<string, number>()
+  const srcLength = src.length
+  for (let idx = 0; src.length > idx; idx++) {
+    if (!Object.hasOwn(src, idx)) {
+      continue
+    }
+    const typeNode = src[idx]
+    if (typeNode.type === 'union') {
+      for (const subTypeNode of typeNode.of) {
+        const hash = hashField(subTypeNode)
+
+        times.set(hash, (times.get(hash) || 0) + 1)
+      }
+    }
+  }
+  if (times.size === 0) {
+    return src
+  }
+
+  const hoisted = new Set<string>()
+  for (let outer = 0; src.length > outer; outer++) {
+    if (!Object.hasOwn(src, outer)) {
+      continue
+    }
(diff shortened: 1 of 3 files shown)


src/typeEvaluator/optimizations.ts
```diff
@@ -62,26 +62,93 @@ export function removeDuplicateTypeNodes(typeNodes: TypeNode[]): TypeNode[] {
   return newTypeNodes
 }
 
+export function hoistDuplicateTypeNodes(src: TypeNode[]): TypeNode[] {
```
judofyr, line 65: Could we add some documentation here? What's the goal here? Is it just about de-duplicating? I'm not sure what "hoist" means in this context. This code also mutates `src` which should be highlighted. Maybe this should not even return a value, but just be a mutable function?

src/typeEvaluator/optimizations.ts
```diff
@@ -62,26 +62,93 @@ export function removeDuplicateTypeNodes(typeNodes: TypeNode[]): TypeNode[] {
...
+  const times = new Map<string, number>()
+  const srcLength = src.length
+  for (let idx = 0; src.length > idx; idx++) {
+    if (!Object.hasOwn(src, idx)) {
+      continue
+    }
```
judofyr, line 71: Shouldn't `idx` always be in bounds?

test/typeEvaluate.test.ts
```diff
@@ -2092,6 +2093,84 @@ t.test('function: references', (t) => {
...
+            value: 3,
+          },
+          {
+            type: 'string',
+            value: 'bar',
+          },
+        ],
+      },
+    ],
+  })
```
judofyr, line 2170: So, the original expression is `("foo" | "bar" | 3) | ("foo" | "baz" | 2 )`, and this asserts that optimizing it should become `"foo" | ("bar" | 3) | ("baz" | 2 )`. However, why don't we want this to become `("foo" | "bar" | 3 | "baz" | 2 )`? Isn't that also equivalent?

judofyr (commented):
I don't quite understand what this is trying to accomplish. Do we have a specific query which is getting incorrectly typed? This says we did something "too aggressively", but it's not clear to me what the bug is here.

sgulseth:
Discussed async, initial idea was fine but I was mixing unions and intersections(which we don't have support for). What we can do is to peek into union of objects, and if all attributes are the same, but rest differs, we can pull the attributes out

PR #232, opened 2024-05-07
fix(typeEvlauator): handle rest on object splat operations

sgulseth:
When splatting an object we didn't map over the rest property, if it's an object we should append the attributes. If `rest` is unknown we should treat the entire object as unknown.

Fixes https://github.com/sanity-io/sanity/issues/6555

### DIFF
--- a/src/typeEvaluator/typeEvaluate.ts
+++ b/src/typeEvaluator/typeEvaluate.ts
@@ -34,7 +34,6 @@ import type {
 import {handleFuncCallNode} from './functions'
 import {optimizeUnions} from './optimizations'
 import {Context, Scope} from './scope'
-import {mapConcrete, nullUnion} from './typeHelpers'
 import type {
   ArrayTypeNode,
   BooleanTypeNode,
@@ -50,7 +49,7 @@ import type {
   UnionTypeNode,
   UnknownTypeNode,
 } from './types'
-import {mapConcrete, nullUnion, resolveInline, type ConcreteTypeNode} from './typeHelpers'
+import {mapConcrete, nullUnion, resolveInline} from './typeHelpers'
 
 const $trace = debug('typeEvaluator:evaluate:trace')
 $trace.log = console.log.bind(console) // eslint-disable-line no-console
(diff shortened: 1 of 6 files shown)


judofyr:
Something looks a bit sketchy.

Here we invoke the mapper multiple times: https://github.com/sanity-io/groq-js/blob/15ec7ac20b00bc3f70ed72c56f8ddcae561930b0/src/typeEvaluator/typeEvaluate.ts#L124-L128

But here we assume that it's only being called once per `name`: https://github.com/sanity-io/groq-js/blob/15ec7ac20b00bc3f70ed72c56f8ddcae561930b0/src/typeEvaluator/typeEvaluate.ts#L165-L167

`mapObjectSplat` also only handles `union` and `object` and ignores `unknown`. It seems that if you now splat something which we can't determine it ends up being ignored? 🤔 

I'm not 100% sure what the right approach here is though. Maybe more usage of `mapConcrete`?

sgulseth:
yess, this should probably be a concrete type with mapConcrete, will fix. Didn't think much about fixing other bugs than the one referenced 😅

sgulseth:
@judofyr did another pass on comparing with mapped concrete type node

src/typeEvaluator/typeEvaluate.ts
```diff
@@ -119,18 +119,40 @@ function handleDerefNode(node: DerefNode, scope: Scope): TypeNode {
...
-) {
-  if (node.type === 'union') {
-    for (const scoped of node.of) {
-      mapObjectSplat(scoped, scope, mapper)
+  mapper: (attribute: ObjectAttribute) => ObjectAttribute,
+): ObjectTypeNode | UnknownTypeNode | NullTypeNode {
+  const attributes: Record<string, ObjectAttribute> = {}
+
+  const concreteNode = mapConcrete(node, scope, (node) => node)
+  if (concreteNode.type === 'object') {
```
judofyr, line 127: Hmm. Doesn't this mean that if this returns `false` we end up returning `{type: 'object', attributes: {}}` below? What about we instead wrap all of this in `mapConcrete`?

```ts
return mapConcrete(node, scope, (node) => {
  if (node.type !== 'object') return {type: 'unknown'}
  const attributes: Record<string, ObjectAttribute> = {}
  // same same…
})
```

sgulseth: I did another pass, moved the handling of the object splats out in its own methods

src/typeEvaluator/typeEvaluate.ts
```diff
@@ -145,44 +230,29 @@ function handleObjectNode(node: ObjectNode, scope: Scope) {
...
-            } else {
-              throw new Error('Unexpected condition')
-            }
-          }
-        })
+      const result = handleObjectConditionalSplatNode(attr, scope)
+      $trace('object.conditional.splat.result %O', result)
+      if (result.type !== 'object') {
+        return result
+      }
```
judofyr, line 250: I don't think this makes sense: If the splatted node is _not_ an object then we return that as the result of the whole object?

In the following query:

```groq
*[_type == "author"][0] {
    _type == "author" => select(age >= 18 => {lastname}, {firstname}),
    _id,
    _type
  }
```

Then we now end up generating:

```js
{
  "type": "union",
  "of": Array [
    Object {
      "type": "object",
      "attributes": Object {
        "firstname": Object {
          "type": "objectAttribute",
          "value": Object {
            "type": "string",
          },
        },
      },
    },
    Object {
      "type": "object",
      "attributes": Object {
        "lastname": Object {
          "type": "objectAttribute",
          "value": Object {
            "type": "string",
          },
        },
      },
    },
    Object {
      "type": "null",
    },
  ],
}
```

Because the "conditional splat" turns into a `union` and thus we hit this case.

src/typeEvaluator/typeEvaluate.ts
```diff
@@ -117,20 +119,103 @@ function handleDerefNode(node: DerefNode, scope: Scope): TypeNode {
...
+    if (mapped.type != 'object') {
+      return mapped
+    }
+    for (const name in mapped.attributes) {
+      if (!mapped.attributes.hasOwnProperty(name)) {
+        continue
+      }
+      attributes[name] = mapped.attributes[name]
+    }
+    return {type: 'object', attributes}
```
judofyr, line 176: This code has just basically takes a `type: "object"` and then makes a copy of it without `rest` or `dereferencesTo`. That doesn't look right to me.

src/typeEvaluator/typeEvaluate.ts
```diff
@@ -117,20 +119,103 @@ function handleDerefNode(node: DerefNode, scope: Scope): TypeNode {
...
+  if (node.rest !== undefined) {
+    // Rest is either an object, inline, or unknown - so mapping the concrete value here and returning it is just to resolve the inline
+    const concreteRest = mapConcrete(node.rest, scope, (rest) => rest)
+
+    // if the rest is unknown the entire object is unknown
+    if (concreteRest.type === 'unknown') {
+      return {type: 'unknown'}
+    }
+    if (concreteRest.type !== 'object') {
+      return {type: 'null'}
```
judofyr, line 148: This means that if `concreteRest.type` is "union" we end up returning null?

judofyr:
It's a bit unclear to me what the intended behavior of handling objects with unions and inlines are at the moment. If there's union in any of `ObjectSplat` there's multiple approaches:

- We can do the the "cross product": E.g. if there's 3 different object splats which all have 3 unions we end up with a top-level union of 9 different alternatives.
- We can choose that this is just always `{type: "unknown"}`.
- We can take the _last_ union of objects, change all of them to have `rest: unknown`. That way we get typings for the "last" set of attributes, but not the "earlier" ones.

Maybe we can add some more fine-grained tests using `overrideTypeForNode` which demonstrates how it will behave in various cases?

judofyr:
How about creating some tests like this (maybe turn it into a table-based one?):

```ts
const nodeWithType = (type: TypeNode) => {
  const expr: ExprNode = {type: 'Value', value: null}
  overrideTypeForNode(expr, type)
  return expr
}

t.test('objects', (t) => {
  // test 1
  typeEvaluate(
    {
      type: 'Object',
      attributes: [
        {type: 'ObjectAttributeValue', name: 'foo', value: nodeWithType({type: 'string'})},
      ],
    },
    [],
  )

  // test 2
  typeEvaluate(
    {
      type: 'Object',
      attributes: [
        {type: 'ObjectAttributeValue', name: 'foo', value: nodeWithType({type: 'string'})},
        {type: 'ObjectAttributeValue', name: 'foo', value: nodeWithType({type: 'number'})},
      ],
    },
    [],
  )

  // test 3
  typeEvaluate(
    {
      type: 'Object',
      attributes: [
        {type: 'ObjectAttributeValue', name: 'foo', value: nodeWithType({type: 'string'})},
        {type: 'ObjectSplat', value: nodeWithType({type: 'unknown'})},
      ],
    },
    [],
  )
})
```

sgulseth:
I've addressed your feedback(100% agree), we should now be handling unions by always using mapConcrete. I also moved resolveInline into a separate function so we can be explicit about resolving inline nodes.

test/typeEvaluate.test.ts
```diff
@@ -2406,6 +2430,800 @@ t.test('function: math::*', (t) => {
...
+      {
+        type: 'object',
+        attributes: {
+          foo: {type: 'objectAttribute', value: {type: 'boolean'}},
+        },
+      },
+      {
+        type: 'unknown',
+      },
+    ),
```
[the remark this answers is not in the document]
sgulseth: We could short circuit so that if any of the returned elements are unknown we return unknown. Guess this could happen if you have a schema where a field is either an object or and array of something, not sure that's very realistic, though it could be during a migration for example. 

If you have multiple attributes in the object, you could simplify the type check by just asserting one value and typescript would assume it's not the unknown object 🤔 

Personally no strong opinions

sgulseth:
Pushed my latest changes, there's optimizations to be done in the object handler itself(probably way too many loops 😅), but if it's correct I can give optimizations another passs

judofyr (approved):
I've mainly focused on the tests and they all look sensible to me.

(I haven't looked much at the implementation. Let me know if there's anything particular there you'd like me to review. Otherwise I'm pretty happy doing a "test-first" approach for this code anyway.)

PR #236, opened 2024-05-15
feat(typeEvaluator): add support for array functions

sgulseth:


judofyr (approved):
I still struggle with understanding the impact of nested `mapConcrete`, but (1) this looks correct and (2) tests are passing.

PR #240, opened 2024-06-17
feat(typeEvaluator): add support for upper() and lower()

sgulseth:


tzhelyazkova (approved):

sgulseth:
@tzhelyazkova pushed a fix to preserve the value if the type has a static value

tzhelyazkova (approved):

PR #241, opened 2024-06-17
feat(typeEvaluator): add support for round()

sgulseth:


test/typeEvaluate.test.ts
```diff
@@ -2072,6 +2072,64 @@ t.test('function: count', (t) => {
...
 })
 
+t.test('function: global::round', (t) => {
+  const query = `*[_type == "author"] {
+    "age": round(age),
+    "number": round(3.14),
+    "precision": round(3.14, 2),
+    "constant": round(3 + 4),
+    "boolean": round(true),
+    "object": round(object)
```
tzhelyazkova, line 2082: why not `"object": round({})` if we're testing an object? 🤔

tzhelyazkova (approved):

test/typeEvaluate.test.ts
sgulseth: yep! Would be the same as your suggestion!

PR #242, opened 2024-06-20
Make sure we are not inside a hidden scope when resolving parent nodes

sgulseth:
When resolving filters we are creating a temporarily hidden scope for each node in the current scope. Since hidden parents are only ignored when creating a new nestedScope(non-hidden) we end up counting the hidden scope.

Fixes sanity-io/sanity#6635


judofyr (approved):
Not a big fan of the `?`s, but I tried to rewrite it in another way and it wasn't any better…

sgulseth:
yeah, agree. Gave it a few attempts myself

PR #248, opened 2024-07-08
chore: upgrade `typescript` and misc KTLO

stipsan:
- [Upgrades TS to 5.5](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/)
- [Removes node 16 test, as it's blocking several dependencies from updating](https://github.com/sanity-io/groq-js/actions/runs/9840041398/job/27163374895?pr=244)
- The matrix uses `current`, `lts/*` and `lts/-1` instead of explicit numbers like `21.x`, so we automatically keep up with the [node release schedule](https://nodejs.org/en/about/previous-releases). 
- Removes the browserslist cronjob (`browserslist.yml`), as renovate already updates the browserslistdb on other PRs.
- Sets up the shared sanity prettier config.


atombender (approved):
Not really my area, but LGTM.

.github/workflows/test.yml
```diff
@@ -34,15 +34,16 @@ jobs:
     # We test on all of the maintained versions: https://nodejs.org/en/about/releases/
     strategy:
       matrix:
-        node-version: ['16.x', '18.x', '20.x', '21.x']
+        node-version: ["lts/-1", "lts/*", "current"]
```
sgulseth, line 37: Why this change? 🤔 Engine states that we should support node 14, so we should test it on node14 as well?

stipsan: Esbuild doesn't support it. There are ways around it but it didn't feel worth the investment for a node version that no longer receives security patches and thus shouldn't be used by anyone in production 😅
At the same time we aren't intentionally breaking node 14 support in userland so it didn't feel right to bump engines 🤷

sgulseth: Right. Is there a way we can still test it with node 14 though? Now we might break the engine support without knowing

PR #249, opened 2024-07-12
Add support for sanity::versionOf()

tzhelyazkova:
Add a new function `sanity::versionOf()` which lists all the stored document IDs that are a version of given published document ID in arg.

### DIFF
--- a/src/typeEvaluator/functions.ts
+++ b/src/typeEvaluator/functions.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-statements */
 import type {FuncCallNode} from '../nodeTypes'
 import {Scope} from './scope'
 import {walk} from './typeEvaluate'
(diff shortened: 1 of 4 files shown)


j33ty:
Tests are failing on prettier missing which is unrelated to this PR.

j33ty:
Pushing an empty commit to re-trigger CI build fixed the issue, however, build rerun on existing commit did not. I am not going to investigate this further, very much feels like an issue with Github Actions.

sgulseth:
I think this was due an issue with node `v22.5.0`, which was fixed in `v22.5.1`, some race condition. `current` isn't marked as required, so should be fine to merge without it passing :)

j33ty:
@sgulseth Good to know. I would also love a review from you here. Let me know if you are missing context around this change.

src/typeEvaluator/functions.ts
```diff
@@ -296,6 +296,9 @@ export function handleFuncCallNode(node: FuncCallNode, scope: Scope): TypeNode {
         })
       })
     }
+    case 'sanity.versionOf': {
+      return {type: 'array', of: {type: 'string'}}
+    }
```
sgulseth, line 301: Ideally we would also check the input to ensure it's a string:

```suggestion
    case 'sanity.versionOf': {
      const typeNode = walk({node: node.args[0], scope})
      return mapConcrete(typeNode, scope, (typeNode) => {
        if (typeNode.type !== 'string') {
          return {type: 'null'}
        }
      	return {type: 'array', of: {type: 'string'}}
      })
    }
```

this can help if you use `versionOf(someField)` and `someField` is not a string

sgulseth (dismissed):

sgulseth (approved):

PR #253, opened 2024-07-26
fix(typeEvaluator): fix in-operator union handling

sgulseth:
We weren't handling union with nulls properly. This reduces over the lhs/rhs on opcall nodes when evaluating conditions

Fixes https://github.com/sanity-io/sanity/issues/6628


tzhelyazkova (approved):
looks good to me but maybe @judofyr also wants to take a look?

sgulseth:
Closing in favor of #256

PR #256, opened 2024-08-18
fix(typeEvaluator): fix in-operator union handling

sgulseth:
Two changes is involved in this pr: 
	* This pr builds on top of #253 which fixes in-operator on union types
	* Removes resolveCondition so that we can reuse types that are walked when we are resolving filters.

### DIFF
--- a/src/typeEvaluator/typeEvaluate.ts
+++ b/src/typeEvaluator/typeEvaluate.ts
@@ -27,6 +27,7 @@ import type {
   SliceNode,
   ValueNode,
 } from '../nodeTypes'
+import {booleanAnd, booleanInterpretationToTypeNode, booleanOr, booleanValue} from './booleans'
 import {handleFuncCallNode} from './functions'
 import {match} from './matching'
 import {optimizeUnions} from './optimizations'
@@ -220,17 +221,17 @@ function handleObjectNode(node: ObjectNode, scope: Scope): TypeNode {
     }
 
     if (attr.type === 'ObjectConditionalSplat') {
-      const condition = booleanValue(walk({node: attr.condition, scope}))
+      const condition = booleanValue(walk({node: attr.condition, scope}), scope)
       $trace('object.conditional.splat.condition %O', condition)
       // condition is never met, skip this attribute
-      if (condition === false) {
+      if (condition.canBeTrue === false) {
         continue
       }
 
       const attributeNode = handleObjectSplatNode(attr, scope)
       $trace('object.conditional.splat.result %O', attributeNode)
       // condition is always met, we can treat this as a normal splat
-      if (condition === true) {
+      if (condition.canBeFalse === false && condition.canBeNull === false) {
         switch (attributeNode.type) {
           case 'object': {
             splatVariants.push([idx, attributeNode])
@@ -1031,28 +1032,9 @@ function handleAndNode(node: AndNode, scope: Scope): TypeNode {
   const right = walk({node: node.right, scope})
   return mapConcrete(left, scope, (lhs) =>
     mapConcrete(right, scope, (rhs) => {
-      if (
-        (lhs.type === 'boolean' && lhs.value === false) ||
-        (rhs.type === 'boolean' && rhs.value === false)
-      ) {
-        return {type: 'boolean', value: false}
-      }
+      const value = booleanAnd(booleanValue(lhs, scope), booleanValue(rhs, scope))
 
-      if (lhs.type !== 'boolean' || rhs.type !== 'boolean') {
-        if (
-          (lhs.type === 'boolean' && lhs.value === undefined) ||
-          (rhs.type === 'boolean' && rhs.value === undefined)
-        ) {
-          return nullUnion({type: 'boolean'})
-        }
-        return {type: 'null'}
-      }
-
-      if (lhs.value === true && rhs.value === true) {
-        return {type: 'boolean', value: true}
-      }
-
-      return {type: 'boolean'}
+      return booleanInterpretationToTypeNode(value)
     }),
   )
 }
@@ -1062,33 +1044,9 @@ function handleOrNode(node: OrNode, scope: Scope): TypeNode {
   const right = walk({node: node.right, scope})
   return mapConcrete(left, scope, (lhs) =>
     mapConcrete(right, scope, (rhs) => {
-      // one of the sides is true the condition is true
-      if (
-        (lhs.type === 'boolean' && lhs.value === true) ||
-        (rhs.type === 'boolean' && rhs.value === true)
-      ) {
-        return {type: 'boolean', value: true}
-      }
+      const value = booleanOr(booleanValue(lhs, scope), booleanValue(rhs, scope))
 
-      // if one of the sides is not a boolean, it's either a null or
-      // a null|boolean if the other side is an undefined boolean
-      if (lhs.type !== 'boolean' || rhs.type !== 'boolean') {
-        if (
-          (lhs.type === 'boolean' && lhs.value === undefined) ||
-          (rhs.type === 'boolean' && rhs.value === undefined)
-        ) {
-          return nullUnion({type: 'boolean'})
-        }
-
-        return {type: 'null'}
-      }
-
-      // both sides are false, the condition is false
-      if (lhs.value === false && rhs.value === false) {
-        return {type: 'boolean', value: false}
-      }
-
-      return {type: 'boolean'}
+      return booleanInterpretationToTypeNode(value)
     }),
   )
 }
@@ -1265,38 +1223,15 @@ function evaluateComparison(
   }
 }
 
-function booleanValue(node: TypeNode): boolean | undefined {
-  // if the node is unknown, we can't match it so we return undefined
-  if (node.type === 'unknown') {
-    return undefined
-  }
-
-  // if the node is a boolean, we can match it, reuse the value
-  if (node.type === 'boolean') {
-    return node.value
-  }
-
-  if (node.type === 'union') {
-    for (const sub of node.of) {
-      const match = booleanValue(sub)
-      if (match !== false) {
-        return match
-      }
-    }
-  }
-
-  return false
-}
-
 // eslint-disable-next-line complexity, max-statements
 function resolveFilter(expr: ExprNode, scope: Scope): UnionTypeNode {
   $trace('resolveFilter.expr %O', expr)
   const filtered = scope.value.of.filter((node) => {
     // create a new scope with the current scopes parent as the parent. It's only a temporary scope since we only want to resolve the condition
-    // and check if the result is "matchable"
-    const cond = walk({node: expr, scope: scope.createHidden([node])})
-    const isMatch = booleanValue(cond)
-    return isMatch === undefined || isMatch === true
+    // and check if the result can be true.
+    const subScope = scope.createHidden([node])
+    const cond = walk({node: expr, scope: subScope})
+    return booleanValue(cond, subScope).canBeTrue
   })
   $trace(
     `resolveFilter ${expr.type === 'OpCall' ? `${expr.type}/${expr.op}` : expr.type} %O`,
--- a/src/typeEvaluator/typeEvaluate.ts
+++ b/src/typeEvaluator/typeEvaluate.ts
@@ -32,7 +32,7 @@ import {handleFuncCallNode} from './functions'
 import {match} from './matching'
 import {optimizeUnions} from './optimizations'
 import {Context, Scope} from './scope'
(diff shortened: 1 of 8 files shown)


src/typeEvaluator/typeHelpers.ts
```diff
@@ -110,6 +111,34 @@ export function mapConcrete(
...
+
+/**
+ * mapUnion extracts a _concrete type_ OR an unknown type from a type node, applies the mapping
+ * function to it and returns. Most notably, this will work through unions
+ * (applying the mapping function for each variant) and inline (resolving the
+ * reference).
+ * This method should _only_ be used if you need to handle unknown types, ie when resolving two sides of an and node, and we don't want to abort if one side is unknown.
+ * In most cases, you should use `mapConcrete` instead.
+ **/
+export function mapUnion(
```
sgulseth, line 136: A bit annoying to need this, but the reason is that when resolving "and" conditions like `_id == $id && _type == "foo"` we don't want to short circuit the mapper when walking `_id == $id`(resolves to `unknown`).

Alternatively we can use mapUnion in `handleOpCallNode` instead, and there do a unknown check so it at least returns a boolean.
```
function handleOpCallNode(node: OpCallNode, scope: Scope): TypeNode {
  $trace('opcall.node %O', node)
  const lhs = walk({node: node.left, scope})
  const rhs = walk({node: node.right, scope})
  return mapUnion(lhs, scope, (left) =>
    // eslint-disable-next-line complexity, max-statements
    mapUnion(rhs, scope, (right) => {
      $trace('opcall.node.concrete "%s" %O', node.op, {left, right})
      if (left.type === 'unknown' || right.type === 'unknown') {
        return {type: 'boolean'}
      }
```


Then we would resolve the correct type for a projection like:
```
{
"foo": $myParameter == 1
}
```

Which currently resolves to `unknown`

test/typeEvaluate.test.ts
```diff
@@ -838,7 +889,7 @@ t.test('values in projection', (t) => {
           type: 'objectAttribute',
           value: {
             type: 'boolean',
-            value: undefined,
+            value: false,
```
sgulseth, line 892: neither `foo` nor `bar` exists on the document, so this will always be false

test/typeEvaluate.test.ts
```diff
@@ -1313,7 +1364,7 @@ t.test('with select, not guaranteed & with fallback', (t) => {
   const query = `*[_type == "author" || _type == "post"] {
         _type,
         "something": select(
-          _id > 5 => _id,
+          _id == "5" => _id,
```
sgulseth, line 1367: _id is a string, so fixed the comparison

test/typeEvaluate.test.ts
```diff
@@ -1672,10 +1723,10 @@ t.test('misc', (t) => {
   const query = `*[]{
       "group": ((3 + 4) * 5),
       "notBool": !false,
-      "notField": !someAttriute,
+      "notField": !someAttribute,
       "unknownParent": ^._id,
       "unknownParent2": ^.^.^.^.^.^.^.^._id,
-      "andWithAttriute": !false && !someAttriute,
+      "andWithAttribute": !false && !someAttribute,
```
sgulseth, line 1729: just a type

src/typeEvaluator/typeEvaluate.ts
```diff
@@ -1047,316 +1149,60 @@ function isPrimitiveTypeNode(node: TypeNode): node is PrimitiveTypeNode {
...
-    }
-
     default: {
-      return undefined
+      throw new Error(`unknown comparison operator ${opcall}`)
     }
   }
 }
 
+function nodeMatch(node: TypeNode): boolean | undefined {
```
judofyr, line 1182: Maybe rename this to `booleanValue(node: TypeNode)` and then clarify that it's intended to be used in the cases where we coerce a value to a boolean (i.e. inside filter + object splat).

Note that in a filter then every other type is actually considered `false`: https://spec.groq.dev/draft/#sel-FARJLCFHFFAACJAuyC. So actually, all of `object`, `number`, `array`, `null` can be considered `false` in this context.

src/typeEvaluator/typeEvaluate.ts
```diff
@@ -960,12 +1037,37 @@ export function walk({node, scope}: {node: ExprNode; scope: Scope}): TypeNode {
...
+      const right = walk({node: node.right, scope})
+      return mapUnion(left, scope, (lhs) =>
+        mapUnion(right, scope, (rhs) => {
+          return {
+            type: 'boolean',
+            value: nodeMatch(lhs) && nodeMatch(rhs),
+          } satisfies BooleanTypeNode
+        }),
+      )
+    }
```
judofyr, line 1051: I think was slightly broken before (because it _always_ returned a `boolean`), but this seems a bit imprecise in how it deals with different types. `&&` and `||` always returns one of `true`, `false`, `null`. So for instance:

- If LHS and RHS are both `boolean` we know that this is always a `boolean`.
- However, if either LHS or RHS are `unknown` then this is actually `boolean | null`.
- Similarly, if LHS or RHS are _known_ to be either `object`, `number`, `array` then it's actually _always_ `null`.

These also quite interestingly compose a bit similarly under `union`.

I think maybe all of this is better expressed with something which turns a `TypeNode` into a type such as:

```ts
type BooleanInterpretation = {
  canBeTrue: boolean
  canBeFalse: boolean
  canBeNull: boolean
}
``` 

`canBeNull` here also includes anything which coerces to `null` in boolean context (i.e. any other type such as `array`).

With the following conversion from TypeNode:

- `boolean` => `{canBeTrue: true, canBeFalse: true, canBeNull: false}`
- `boolean | null` => `{canBeTrue: true, canBeFalse: true, canBeNull: true}`
- `unknown` => `{canBeTrue: true, canBeFalse: true, canBeNull: true}`

This at least makes it easier to reason about the logic.

E.g. for `And` we know that if either side is `canBeTrue: true, …: false` then it short-circuits and returns the other side. And if `canBeTrue: false, …` then everything becomes `false` actually.

TL;DR as I'm writing this: This `BooleanInterpretation` is maybe just a more precise version of `nodeMatch`.

src/typeEvaluator/typeHelpers.ts
judofyr: And similarly: Comparisons operators are always `boolean | null`. We should _always_ be able to fallback to that.

judofyr (commented):
Having evaluate type tests for `And` and `Or` will probably uncover potential bugs around the handling.

```ts
t.test(`And`, async (t) => {
  subtestBinary({
    t,
    variants1: trivialVariant,
    variants2: trivialVariant,
    build: (left, right) => ({type: 'And', left, right})
  })
})

t.test(`Or`, async (t) => {
  subtestBinary({
    t,
    variants1: trivialVariant,
    variants2: trivialVariant,
    build: (left, right) => ({type: 'Or', left, right})
  })
})
```

src/typeEvaluator/typeEvaluate.ts
```diff
@@ -460,55 +457,149 @@ function handleOpCallNode(node: OpCallNode, scope: Scope): TypeNode {
...
+        }
+        case 'in': {
+          if (left.type === 'unknown' || right.type === 'unknown') {
+            return {type: 'unknown'}
+          }
+          if (right.type !== 'array') {
+            // Special case for global::path, since it can be used with in operator, but the type returned otherwise is a string
+            if (isFuncCall(node.right, 'global::path')) {
+              return {type: 'boolean'}
+            }
```
sgulseth, line 549: A bit hacky, but just needed to short circuit the operator if used with path.

Another approach, that might come in a follow-up PR is to put the scope into `filterMode`, and we can then use that to determine what typenode to return in the function. This could also then be used to rewrite the scope when using literal comparisons or the `defined` method

src/typeEvaluator/typeHelpers.ts
sgulseth: Change this into `function handleOpCallNode` handling unknowns with mapUnion, and `==`/`!=` will now always return a `boolean`, while the other operators short circuits to `unknown` if an `unknown` is given.

src/typeEvaluator/typeEvaluate.ts
```diff
@@ -905,6 +1026,70 @@ function handleEverythingNode(_: EverythingNode, scope: Scope): TypeNode {
...
+        if (leftValue.canBeFalse || rightValue.canBeFalse) {
+          if (leftValue.canBeNull || rightValue.canBeNull) {
+            return nullUnion({type: 'boolean'})
+          }
+
+          return {type: 'boolean'}
+        }
+
+        return {type: 'boolean', value: true}
+      }
```
judofyr, line 1078: Isn't this wrong if both values are `[canBeTrue, canBeNull]` and `[canBeTrue, canBeNull]`? 🤔 

In which case the actual type should be `nullIUnion({type: 'boolean'})`, but here it would be detected as `{type: 'boolean', value: true}`?

I think potentially this can be easier expressed by defining the logic on BooleanInterpretation:

```js
function booleanOr(left: BooleanInterpretation, right: BooleanInterpretation): BooleanInterpretation {
  // One side is true => return `true`
  if (left.canBeTrue && !left.canBeFalse && !left.canBeNull) return left
  if (right.canBeTrue && !right.canBeFalse && !right.canBeNull) return right

  // One side is false => return the other side
  if (!left.canBeTrue && left.canBeFalse && !left.canBeNull) return right
  if (!right.canBeTrue && right.canBeFalse && !right.canBeNull) return left

  // Guaranteed to be boolean at least:
  if (!left.canBeNull && !right.canBeNull) return {canBeTrue: true, canBeFalse: true, canBeNull: false}

  // Fallback to the most permissive:
  return {canBeTrue: true, canBeFalse: true, canBeNull: true}
}
```

I think this covers the main cases.

And then `and` becomes quite symmetric:

```js
function booleanAnd(left: BooleanInterpretation, right: BooleanInterpretation): BooleanInterpretation {
  // One side is true => return that side
  if (left.canBeTrue && !left.canBeFalse && !left.canBeNull) return right
  if (right.canBeTrue && !right.canBeFalse && !right.canBeNull) return left

  // One side is false => return `false`
  if (!left.canBeTrue && left.canBeFalse && !left.canBeNull) return left
  if (!right.canBeTrue && right.canBeFalse && !right.canBeNull) return right

  // Guaranteed to be boolean at least:
  if (!left.canBeNull && !right.canBeNull) return {canBeTrue: true, canBeFalse: true, canBeNull: false}

  // Fallback to the most permissive:
  return {canBeTrue: true, canBeFalse: true, canBeNull: true}
}
```

src/typeEvaluator/typeHelpers.ts
```diff
@@ -110,6 +112,34 @@ export function mapConcrete(
   mapper: (node: ConcreteTypeNode) => TypeNode,
   mergeUnions: (nodes: TypeNode[]) => TypeNode = (nodes) =>
     optimizeUnions({type: 'union', of: nodes}),
+): TypeNode {
+  return mapUnion(
```
judofyr, line 116: Actually, how about we take the `mapUnion` here and just call it `mapConcrete`? So `mapConcrete` will call the function with either a concrete value _or_ `unknown`. And then it's up to the caller to decide how to deal with `unknown`. The more I think I about it, the less it seems like "`unknown` input implies `unknown` output" is a good assumption. In many scenarios we know way more about the output type (e.g. `round()` is either way `null` or `number` even if the input is `unknown`.

src/typeEvaluator/typeEvaluate.ts
```diff
@@ -866,28 +977,38 @@ function handleParentNode({n}: ParentNode, scope: Scope): TypeNode {
 
 function handleNotNode(node: NotNode, scope: Scope): TypeNode {
   const base = walk({node: node.base, scope})
-  if (base.type === 'boolean' && base.value !== undefined) {
-    return {type: 'boolean', value: base.value === false}
-  }
-  return {type: 'boolean'}
+  return mapConcrete(base, scope, (base) => {
```
judofyr, line 980: (Related to my feedback on `mapUnion`/`mapConcrete`: If we make this `unknown`-aware we can actually return `nullUnion({type: 'boolean'})` in this case)

judofyr (changes requested):
One proper request: The handling of `null` + `true` seems broken in And/Or. I think this would have been covered by adding a `typeEvaluateCompare` test where we add `union({type: 'null'}, …)`. Currently we _only_ union by `unknown`. It's a bit tricky to add this test without exploding all of the variants though. We might need more fine-grained per-op variations…

One suggestion which can be done separately from this PR: Don't distinguish between `mapConcrete` and `mapUnion`, but _always_ use one (I propose keeping the name `mapConcrete`) which also requires you to handle `unknown`.

sgulseth:
Fixed in #260

PR #257, opened 2024-08-21
Add array intersects function

j33ty:
Add a new function `array::intersects()` which returns true/false if some of the strings in the argument array intersect.


src/evaluator/functions.ts
```diff
@@ -19,6 +19,7 @@ import {portableTextContent} from './pt'
 import {Scope} from './scope'
 import {evaluateScore} from './scoring'
 import type {Executor} from './types'
+import {isEqual} from './equality.ts'
```
atombender, line 22:
```suggestion
import {isEqual} from './equality'
```

src/evaluator/functions.ts
```diff
@@ -392,6 +393,31 @@ array['unique'] = async function (args, scope, execute) {
...
+    return NULL_VALUE
+  }
+
+  const arr2 = await execute(args[1], scope)
+  if (!arr2.isArray()) {
+    return NULL_VALUE
+  }
+
+  for (const v1 of await arr1.get()) {
+    for (const v2 of await arr2.get()) {
```
atombender, line 410: In other places we use `for await (const elem of arr)`, should we follow that pattern?

j33ty: Sure.

j33ty:
Tests are failing here because groq-test-suite is using `geo` data type in `array::intersects()` function without gating these tests behind `geoFunctions` feature. GROQ JS does not seem to support geo data type and geo functions.

1. https://github.com/sanity-io/groq-test-suite/pull/80/files#diff-9281d464e6eb3b49a5eb063cdc52d17e81f1f2900e73d1358396115e0bf7e1c0R16
2. https://github.com/sanity-io/groq-js/blob/897de24b851f84683239a57aa872c9b507416a1c/test/generate.js#L11
3. https://github.com/sanity-io/groq-test-suite/blob/main/test/function/geo/geo.yml#L3

atombender (approved):
LGTM!

atombender:
> without gating these tests behind geoFunctions feature

That's an oversight, yes.

j33ty:
PR https://github.com/sanity-io/groq-test-suite/pull/85 fixes the `geoFunctions` feature on geo data type.

atombender (approved):
LGTM!

PR #260, opened 2024-08-23
Remove resolve condition, use walk to determine conditions

sgulseth:
Two changes included in this PR:
	* Removes resolveCondition so that we can reuse types that are walked when we are resolving filters.
	* Introduces BooleanInterpretation that can be used to determine boolean conditions


judofyr (approved):
LGTM. I've focused mainly on:

* Correctness of the boolean interpretation logic.
* That the test cases look sensible.

Let's improve the implementation further later by adding more tests.

PR #261, opened 2024-08-23
Handle unknowns

sgulseth:
This PR changes so that mapConcrete also returns `unknown`, then it's up to the caller to handle these cases. Most notable this let us do unions of `null`/expectedType when unknown given


judofyr (dismissed):
Looks good. Interesting that none of tests had to be changed. I guess we don't have so much _specific_ tests for these `unknown` cases.

PR #264, opened 2024-11-19
fix: pipe functions in projections

okj579:
This fixes pipe functions inside object projections, like the query below:

```js
const input = `*[_type == "user"] {
    name,
    refs[]-> | order(name)
}`;
const dataset = [
    {_id: '1', _type: 'user', name: 'Michael', refs: [{_ref:'1'}, {_ref:'2'}]},
    {_id: '2', _type: 'company', name: 'Bluth Company'},
];
```

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -833,6 +833,7 @@ function extractPropertyKey(node: ExprNode): string {
   }
 
   if (
+    node.type === 'PipeFuncCall' ||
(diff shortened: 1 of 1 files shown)


sgulseth (approved):

sgulseth:
Thanks! Sorry it took so much time, wanted to make sure we had test coverage for this case

PR #268, opened 2024-12-04
fix: ensure coalesce handles optional/unknown/null properly

sgulseth:
We weren't really handling coalesce well:
1. We were pushing `null` into the list of possible types
2. We weren't handling unknowns
3. Even if we encountered a node that could never be null due to non-optional or no nulls in union we were still continuing

Opted to remove the snapshot and rather be explicit

### DIFF
--- a/src/typeEvaluator/functions.ts
+++ b/src/typeEvaluator/functions.ts
@@ -1,5 +1,6 @@
 /* eslint-disable max-statements */
 import type {FuncCallNode} from '../nodeTypes'
+import {optimizeUnions} from './optimizations'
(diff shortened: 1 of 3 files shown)


judofyr (approved):
Tests look good and then I'm happy 🥳  Haven't looked deeply into the logic itself, but if we think there's still some bugs there I think we should rather invest in getting the `typeEvaluateCompare` tests on `coalesce` since we can then test a lot of combinations.

sgulseth:
yess, agree on the compare 🙃 Soon!

PR #272, opened 2025-02-17
feat(evaluator): add `evaluateSync`

ricokahler:
work in progress


judofyr (commented):
Looking over the code now I think we wouldn't actually even need an `evaluateSync`-mode, but could instead say that it all depends on the implementation of the `dataset`: If `dataset` is a synchronous value I think we can guarantee that the result will also be a synchronous value.

Shouldn't we now be able to use the same `co`-trick, but generate a value which represents an array behind the scenes, and still be incremental? Basically change `StreamValue` to take in a similar coroutine, and then it can also act in synchronous mode if needed? 🤔

ricokahler:
> Looking over the code now I think we wouldn't actually even need an `evaluateSync`-mode, but could instead say that it all depends on the implementation of the `dataset`: If `dataset` is a synchronous value I think we can guarantee that the result will also be a synchronous value.

I like this. we could also add some function overloads to the types to that assert a value is synchronous if a synchronous iterator was passed as the dataset or not, however to fully realize this idea, we may want to take a step further and return an async iterator as the result ourselves. this may not be very practical though as a lot of groq operations require exhausting the iterable so hard to say how much benefit there is to it but it's in the spirit of it.

I also figured part of the reason why this lib is async is to attempt to break up the execution to allow queries on large datasets to not block the main thread. I see that StreamValue awaits nextTick before yielding. If we change this default behavior, does that constitute a break change and if so, are we willing to do that?

> Shouldn't we now be able to use the same `co`-trick, but generate a value which represents an array behind the scenes, and still be incremental? Basically change `StreamValue` to take in a similar coroutine, and then it can also act in synchronous mode if needed? 🤔 

The co trick works for unwrapping promises but it does not normalize between async and sync iteration. (`for await` vs just `for`). It's possible to normalize these if the Values themselves had higher level operations that returned a promise or not (i did this by adding `first()` to both StaticValue and StreamValue) but otherwise, two separate implementations for sync and async are needed.

for operations that require collecting all the values before emitting (e.g. scoring, unique), there shouldn't really be a disadvantage in collecting to an array first and processing synchronously but for certain functions (intersects) the result could be returned without exhausting the iterable which is very appealing for async iterables.

some specific questions:
1. how important is breaking up execution (i.e. await nextTick()) to avoid blocking the main thread while processing in "an async mode"? this could be toggled regardless if the input is a synchronous iterable or not and has its own benefits and drawbacks. this is the only place i see where we are intentionally deferring execution
2. if we change the behavior to synchronous execution when passed in a synchronous iterable (an array) and also remove the await nextTick() delay, is this considered a breaking change? if so, are we willing to move forward with that?
3. how important is proper async iteration (possible to finish execution without exhausting the input async iterable)? if this is important, then we'd most likely need to implement two separate implementations for certain operations and have some sort of `mode` flag brought through the entire evaluation

If 1 is not important, then we can remove this. If we remove the await nextTick, the only reason to support async execution is to benefit from the lazy execution when providing an async iterable. this might be beneficial if say streaming in an NDJSON file from
the network. However this makes the answer to 3 be very important. If this is important, then are we okay with a `mode` flag being passed throughout the evaluator like it currently is in this draft implementation?

If the answer to 1 and 3 are not very important, then I recommend we only have a synchronous evaluator and we remove all the async logic but that's also a large change. Is that change breaking? Is that the direction we want to move forward with?

If 1 is important and 3 is not important then we can support and async and sync mode without needing to pass a "mode" flag around but the async iterable will always be collected and processed synchronously. we can still yield between operations though to break up execution but yielding between items in arrays seems contradictory give the next tick can't continue until all of it is collected.

PR #275, opened 2025-02-28
fix(functions): correct behaviour of partOfRelease and versionsOf

rneatherway:
These were looking for the `_version` attribute, which was deprecated
in favour of the `versions.` id prefix.


src/evaluator/functions.ts
```diff
@@ -460,22 +460,25 @@ sanity['versionOf'] = async function (args, scope, execute) {
   if (value.type !== 'string') return NULL_VALUE
   const baseId = value.data
 
-  // All the document are a version of the given ID if:
-  //  1. Document ID is of the form bundleId.documentGroupId
-  //  2. And, they have a field called _version which is an object.
+  // All the document are a version of document 'bar' if its document ID is of
```
tzhelyazkova, line 463:
```suggestion
  // All the documents are a version of document 'bar' if its document ID is of
```

tzhelyazkova (approved):

tzhelyazkova (approved):

PR #278, opened 2025-03-08
fix(typeEvaluator): avoid recalculating typeNode hash

sgulseth:
Slow query+schema is now takes <1s to evaluate 🙃


src/typeEvaluator/optimizations.ts
```diff
@@ -8,53 +8,71 @@ function typeNodesSorter(a: TypeNode, b: TypeNode): number {
   return compare(hashField(a), hashField(b))
 }
 
+const hashCache = new WeakMap<TypeNode, string>()
+
 export function hashField(field: TypeNode): string {
```
judofyr, line 13: How about moving the caching into a wrapper function instead?

```ts
const hashCache = new WeakMap<TypeNode, string>()

export function hashField(field: TypeNode): string {
  if (hashCache.has(field)) {
    return hashCache.get(field)
  }

  const res = hashFieldUncached(field)
  hashCache.set(field, res)
  return res
}

export function hashFieldUncached(field: TypeNode): string {
 // Same code (which calls `hashField`)
}
```

sgulseth: Done!

PR #280, opened 2025-03-31
chore: update ESLint configuration and dependencies; restructure repo

ricokahler:
- Enhanced ESLint configuration by adding new rules and plugins, including `eslint-plugin-import-x` and `eslint-plugin-tsdoc`, to improve code quality and documentation.
- Updated `package.json` to reflect the addition of new dependencies and removed obsolete entries.
- Refactored the project structure by deleting the `src/1.ts` file and reorganizing exports to streamline the codebase.
- Introduced a new `markProcessor` and `parser` module to improve parsing capabilities and maintainability.

PR #279, opened 2025-03-31
chore!: bump minimum Node.js engine to 18.6.0

ricokahler:
BREAKING CHANGE: The minimum Node.js engine requirement has been bumped from >=14 to >=18.6.0 to match tap's minimum engine requirement. This ensures compatibility with our test suite.

- Update tap from 16.3.10 to 21.1.0
- Update various dev dependencies
- Clean up linting rules and improve code quality

PR #281, opened 2025-04-01
feat: synchronous evaluator

ricokahler:
**feat: Make `evaluate` synchronous and introduce `toJS`**

### Motivation

The SDK team utilizes `groq-js` to calculate user permissions based on GROQ filters applied to documents. This calculation needs to happen synchronously within the same frame where the document data is available, essentially making permissions a direct function of the current document state. The previous asynchronous nature of the `evaluate` function (returning a `StreamValue` with an async `.get()` method) made this pattern challenging to implement reliably.

### Solution

This PR fundamentally rewrites the core `evaluate` function in `groq-js` to be **synchronous**.

1.  **Synchronous `evaluate`:** The `evaluate` function now directly returns an internal `Value` representation without involving Promises.
2.  **Introduce `toJS`:** Since `evaluate` returns an internal representation (which may include iterators or custom classes like `DateTime`), a new helper function `toJS(value)` has been introduced. This function eagerly converts the internal `Value` into a standard, plain JavaScript object, array, or primitive, resolving iterators and converting custom types (like `DateTime` to ISO 8601 strings).
3.  **Simplified Implementation:** The internal evaluator logic has been refactored and simplified, aiming for a more idiomatic JavaScript approach.
4.  **Polyfills for Compatibility:** To maintain compatibility with environments lacking certain newer JavaScript features (like `Iterator.prototype.from`), `core-js-pure` has been added as a dependency, and relevant polyfills are used internally. While this adds some code complexity, it's a trade-off for broader usability. This dependency could potentially be removed in the future as target environments evolve.

### Key Changes

*   `evaluate` function is now synchronous.
*   Removed the asynchronous `value.get()` method previously used to resolve results.
*   Added a new exported function `toJS(value)` for converting evaluation results to plain JavaScript.
*   Updated `README.md` and `API.md` to reflect the new synchronous `evaluate` and `toJS` usage pattern.
*   Removed internal `StreamValue` implementation; now likely uses native iterators where applicable.
*   Internal `DateTime` class added/modified for date handling.
*   Added `core-js-pure` dependency and integrated necessary polyfills.
*   Significant refactoring of internal evaluation logic, operators, and functions to support the synchronous model.

### Context & Collaboration

This change mostly aligns with discussions and a proof-of-concept (#273) reviewed by @judofyr. @judofyr agreed that a synchronous evaluator makes sense for a potential `groq-js` v2, allowing for an `evaluateAsync` to potentially be reintroduced later if needed.

> I think this makes a lot of sense for a groq-js v2. We can still in a minor release introduce an `evaluateAsync` later on.
> 
> > It seems to be pretty doable but handling dates requires checking a string to see if it's a date first and that may be problematic. At the current state of this PR, the date handling is close but still wrong.
> 
> I think we might need to keep our own `Date` wrapper, but there's a few things we can do:
> 
> 1. We can introduce an evaluation option called `normalize` (which defaults to true) which at the end traverses the object deeply and gets of rid any GROQ-specific types.
> 2. We can implement `toJSON()` on our `Date` wrapper so that it's possible to run `JSON.stringify(…)` on it directly.
> 
> Let's also be way more precise (both in the API and preferably in the types) around what types we actually handle in the evaluator. We shouldn't need to use `any` / `unknown` anywhere.

- The DateTime wrapper was kept ✅
- _We can introduce an evaluation option called `normalize`._ I did not do this yet. Do you think it's still worth adding?
- _We can implement `toJSON()` on our `Date`_ This was done. ✅
- _Let's also be way more precise (both in the API and preferably in the types)._ This is mostly the case but I may have miss some functions that could have more strict return types. The `Value` type is meant to realize this vision.

**API Changes Summary:**

*   `evaluate(tree, options)`: Now returns a synchronous internal `Value`.
*   `value.get()`: Removed.
*   `toJS(value)`: New function to get plain JS data from the result of `evaluate`.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,82 +1,213 @@
 # API
 
-GROQ-JS exposes three functions:
+`groq-js` provides functions to parse and evaluate GROQ queries in JavaScript environments.
+
+## Core Functions
+
+The primary functions you'll likely use are:
 
 ```typescript
-import {parse, evaluate, typeEvaluate} from 'groq-js'
+import {parse, evaluate, toJS} from 'groq-js'
+
+// Parse a GROQ query string into an Abstract Syntax Tree (AST)
+const tree = parse('*[_type == "product"]{name, price}')
+
+// Evaluate the AST against a dataset
+const dataset = [
+  /* ... your data ... */
+]
+const value = evaluate(tree, {dataset})
+
+// Convert the internal result representation to a standard JavaScript value
+const result = toJS(value)
+
+console.log(result) // Output: Plain JavaScript array/object/value
 ```
 
-- [`parse`](#parse)
-- [`evaluate`](#evaluate)
-- [`typeEvaluate`](#typeEvaluate)
+- [`parse`](#parse): Parses a GROQ query string into an AST node (`ExprNode`).
+- [`evaluate`](#evaluate): Evaluates an AST node against a dataset and context, returning an internal `Value` representation.
+- [`toJS`](#tojs): Converts the internal `Value` representation into a plain JavaScript value.
+- [`typeEvaluate`](#typeevaluate): (Beta) Evaluates the _type_ that a query would return based on a schema.
+
+## Other Exports
+
+The library also exports various types, errors, and helper functions:
+
+```typescript
+import {
+  // Core Functions (already shown)
+  // parse, evaluate, toJS, typeEvaluate
+
+  // Errors
+  GroqSyntaxError, // Thrown by parse for syntax issues
+  GroqQueryError, // Thrown during evaluation for runtime issues
+
+  // Type Evaluation Helpers & Types (used with typeEvaluate)
+  createReferenceTypeNode,
+  SchemaType, // Your schema definition type
+  TypeNode, // The resulting type representation
+  // ... specific TypeNode subtypes (ArrayTypeNode, StringTypeNode, etc.)
+  // ... specific SchemaType subtypes (DocumentSchemaType, etc.)
+
+  // Core Types
+  ExprNode, // Base type for AST nodes from parse()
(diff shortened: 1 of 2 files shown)


ricokahler:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sanity-io/groq-js/281?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#281** <a href="https://app.graphite.dev/github/pr/sanity-io/groq-js/281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sanity-io/groq-js/281?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#280** <a href="https://app.graphite.dev/github/pr/sanity-io/groq-js/280?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#279** <a href="https://app.graphite.dev/github/pr/sanity-io/groq-js/279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `v2`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

ldmichae:
Very interested in this. Wanting to use groq-js in my corporate projects, but the async nature of the library makes implementation require too many updates to existing codebases. Thanks for doing this work @ricokahler 🎉

PR #289, opened 2025-07-22
feat: ast node text serialization for pretty printing

runeb:
### Description

This PR introduces a node serializer which can be used as a pretty printer. The serializer takes parsed GROQ AST nodes and outputs clean, consistently formatted GROQ query strings.

The serializer implements the complete GROQ specification for string literals, operators, objects, arrays, and all query constructs. It handles proper indentation for nested structures, consistent spacing around operators, and correct escaping of string literals according to the GROQ spec.

Key features:
- Serializes all GROQ node types including filters, projections, function calls, and pipe operations
- Proper string escaping for all control characters defined in the GROQ spec (quotes, backslashes, newlines, tabs, etc.)
- Consistent indentation and spacing for nested objects and arrays
- Preserves explicit parentheses from the original AST

### What to review

Reviewers should focus on:

1. **Core serializer implementation** in `src/serializer/` - verify that all GROQ node types are handled correctly
2. **String escaping logic** in `formatValue()` - ensure it matches the GROQ specification for string literals
3. **Test coverage** in `test/serializer.test.ts` - added tests covering many serializing scenarios
4. **Round-trip behavior** - verify that `parse(serialize(parse(query)))` produces consistent results
5. **Integration** - the serializer is exported from a new beta entry point

Test the formatter by:
- Importing `{serialize}` from `groq-js/beta`
- Parsing a GROQ query with `parse()`
- Formatting the result with `serialize(tree)`
- Verifying the output is valid, readable GROQ

### Testing

Added test coverage including:

- Basic value serialization (strings, numbers, booleans, null)
- String escaping for all GROQ spec control characters (\n, \r, \t, \f, \b, \", \\)
- Object and array serializing with proper indentation
- Complex nested query structures
- Operator serializing and spacing
- Function calls and pipe operations
- Round-trip parsing validation
- Property name shorthand vs explicit key serializing

There may be parts of the spec uncovered by current serializing tests!


src/formatter/nodeFormatters.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  current(): string {
+    return this.indentString.repeat(this.currentIndent)
+  }
+
+  newLine(): string {
+    return `\n${this.current()}`
+  }
+}
+export class NodeFormatter {
```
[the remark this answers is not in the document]
runeb: Excellent input and I agree, its working on AST nodes to serialize them to text currently, but I think the API might not be right at the moment. It could be better to have a `format(queryString: string): string` which takes queries and returns pretty printed queries. Then it would more likely be called either a formatter, or perhaps a printer?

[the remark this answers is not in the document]
runeb: But we might want to leave room for other kinds of serializers, for example a compact printer or obfuscator or... I can change the name here to serializer and we can instead consider the user facing API change separately

src/formatter/nodeFormatters.ts
```diff
@@ -0,0 +1,393 @@
...
+  private indent: IndentationManager
+
+  constructor(indentString: string) {
+    this.indent = new IndentationManager(indentString)
+  }
+
+  format(node: ExprNode): string {
+    return this.formatNode(node)
+  }
```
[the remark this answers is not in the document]
runeb: Fixed by creating it when serializing

src/serializer/nodeSerializers.ts
```diff
@@ -0,0 +1,402 @@
...
+
+        // Special handling for dereferencing projections: author->{...}
+        // Only use shorthand if property name matches the field being dereferenced
+        if (
+          attr.value.type === 'Projection' &&
+          attr.value.base.type === 'Deref' &&
+          attr.value.base.base.type === 'AccessAttribute' &&
+          attr.value.base.base.name === attr.name
+        ) {
+          const space = ' '
```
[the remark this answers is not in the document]
sgulseth: can you just add a space to the string? :P Why assign it to a variable? 😅

src/serializer/nodeSerializers.ts
```diff
@@ -0,0 +1,402 @@
...
+      case 'ObjectSplat':
+        // For object spread, omit 'This' (@) since it's implicit
+        if (attr.value.type === 'This') {
+          return '...'
+        }
+        return `...${this.serializeNode(attr.value)}`
+      case 'ObjectConditionalSplat':
+        return `${this.serializeNode(attr.condition)} => ${this.serializeNode(attr.value)}`
+      default:
+        throw new Error(`Unknown object attribute type: ${(attr as ObjectAttributeNode).type}`)
```
sgulseth, line 243: Let's use the same trick here
```suggestion
				// @ts-expect-error handle all cases
        throw new Error(`Unknown object attribute type: ${attr.type}`)
```

src/serializer/nodeSerializers.ts
```diff
@@ -0,0 +1,402 @@
...
+    return this.serializeBinaryOp(node.left, node.op, node.right)
+  }
+
+  private serializeBinaryOp(left: ExprNode, op: string, right: ExprNode): string {
+    const leftStr = this.serializeWithParens(left)
+    const rightStr = this.serializeWithParens(right)
+
+    if (op === ':') {
+      return `${leftStr}: ${rightStr}`
+    }
```
sgulseth, line 285: binary op shouldn't be `:`? 🤔 
```suggestion
```

src/serializer/nodeSerializers.ts
```diff
@@ -0,0 +1,402 @@
...
+      case 'Tuple':
+        return this.serializeTuple(node)
+      case 'Map':
+        return this.serializeMap(node)
+      case 'FlatMap':
+        return this.serializeFlatMap(node)
+      case 'Context':
+        return this.serializeContext(node)
+      case 'Selector':
+        return '<selector>'
```
sgulseth, line 135: better to throw an error than to generate invalid groq:
```suggestion
        throw new Error('Can't serialize Selector')
```

test/serializer.test.ts
```diff
@@ -0,0 +1,535 @@
+import t from 'tap'
```
sgulseth, line 1: Some tests here are testing that X in gives X out. Could we change it to something like:
```
const queries: string[] = [
	"*",
	"@",
....
]

for (const query of queries) {
	t.test('query: ' + query, async (t) => {
    const expected = parse(query)
    const result = serializeString(tree)
    t.strictSame(parse(result), expected)
  })
}
```

makes it easier to add a bunch of queries to test and that we generate the same identical queries, kinda similar to round-trip only that we compare the ast and not the generated string

src/serializer/nodeSerializers.ts
```diff
@@ -0,0 +1,402 @@
...
+      return `"${value
+        .replace(/\\/g, '\\\\')
+        .replace(/"/g, '\\"')
+        .replace(new RegExp(backspace, 'g'), '\\b')
+        .replace(/\f/g, '\\f')
+        .replace(/\n/g, '\\n')
+        .replace(/\r/g, '\\r')
+        .replace(/\t/g, '\\t')}"`
+    }
+    return String(value)
```
sgulseth, line 156: `JSON.stringify` should return a compatible string without regex foo:
```suggestion
    return JSON.stringify(node.value)
```

src/serializer/nodeSerializers.ts
```diff
@@ -0,0 +1,402 @@
...
+
+  private serializeContext(node: ContextNode): string {
+    return `${node.key}()`
+  }
+
+  private serializeWithParens(node: ExprNode): string {
+    // Only preserve explicit parentheses from the original AST (Group nodes)
+    // Don't add any new parentheses based on precedence
+    return this.serializeNode(node)
+  }
```
sgulseth, line 401: This seems redundant?
```suggestion
```

src/beta.ts
```diff
@@ -0,0 +1,16 @@
+// Beta/experimental features - API subject to change
```
sgulseth, line 1: Nit/Question: Is `experimental` a better name? `groq-js/experimnetal` versus `groq-js/beta`? 🤔

src/serializer/nodeSerializers.ts
```diff
@@ -0,0 +1,402 @@
...
+    return `(${members.join(', ')})`
+  }
+
+  private serializeMap(node: MapNode): string {
+    // Map operations - handle projections specially
+    if (node.expr.type === 'Projection') {
+      // This is a projection like *[condition] {...} or chained projections
+      // Add space only in pretty mode
+      const space = ' '
+      return this.serializeWithParens(node.base) + space + this.serializeNode(node.expr)
```
sgulseth, line 384:
```suggestion
      return this.serializeWithParens(node.base) + ' ' + this.serializeNode(node.expr)
```

src/serializer/nodeSerializers.ts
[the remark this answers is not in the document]
runeb: Hah, yes this is absolutely not necessary after removing the configurable spacing!

test/serializer.test.ts
runeb: Adding a collection of groq queries derived from our documentation

runeb:
Changing to draft while refactoring beta/experimental

src/beta.ts
runeb: Renamed to experimental and re-export `parse` from there so library users only need to integrate to one entrypoint. Also solves problems with incompatible types from the different entrypoints

Issue #141, opened 2023-08-28
geo functions


saiichihashimoto:
The [groq spec](https://sanity-io.github.io/GROQ/GROQ-1.revision1/#sec-Geography-Extension) and the [sanity docs](https://www.sanity.io/docs/groq-functions#138f4451c0aa) both mention the `geo()` function and namespace, but groq-js doesn't support it yet.

judofyr:
There's currently no plan of implementing this in the short-term since they are quite complicated. This probably needs to be done in an opt-in way since we'd have to depend on a geo library. Open for suggestions!

PR #290, opened 2025-07-27
Add geo namespace with noop

LudvigHz:
### Description

The geo function is part of the spec as an extension, but is not included as part of groq-js. This makes groq-js non-compliant in accordance with the spec with extensions. This is especially problematic in places where groq-js is relied upon to safely parse queries but not evaluate anything, eg. sanity codegen.

As functions in the text namespace are already implemented with noops to not break the parser, I have done the same for all functions in the geo namespace.

Although not a complete implementation, this is miles better than today, as the parser actually works on groq queries used in production code today. This means that projects using these features finally can at least somewhat rely on tools requiring groq-js to parse queries.

### Testing

I have tested this change together with `@sanity/cli` running `sanity typegen generate` on a mid-sized project using geo functions a lot. This change makes the codegen successfully parse and generate valid (although "incorrect") types for queries containing any use of the `geo` functions.

Related to #141

### DIFF
--- a/src/evaluator/functions.ts
+++ b/src/evaluator/functions.ts
@@ -768,6 +768,20 @@ text['query'] = () => {
 }
 text['query'].arity = 1
 
+const geo: FunctionSet = {}
+geo['latLng'] = () => {
+  throw new Error('not implemented')
+}
+geo['containes'] = () => {
+  throw new Error('not implemented')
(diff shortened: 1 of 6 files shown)


sgulseth:
Thanks! We can iterate on this :)

PR #292, opened 2025-08-06
feat: Add parser error messages

snocorp:
### Description

When the parser encounters a syntax error, we now return a message indicating what was wrong. This should be helpful in diagnosing problems and debugging issues.

### What to review

Ensure all errors correctly describe the problem and are grammatically correct and understandable.

### Testing

Unit tests should verify each type of error returns the expected error message.

### DIFF
--- a/src/rawParser.d.ts
+++ b/src/rawParser.d.ts
@@ -13,5 +13,5 @@ export declare function parse(input: string):
   | {
       type: 'error'
       position: number
-      error: string
+      message: string
     }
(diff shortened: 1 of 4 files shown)


src/rawParser.d.ts
```diff
@@ -13,4 +13,5 @@ export declare function parse(input: string):
   | {
       type: 'error'
       position: number
+      error: string
```
sgulseth, line 16: bike shedding: should this be `message`? 😅

sgulseth (approved):
nice!

src/rawParser.d.ts
snocorp: Now that you say that I definitely prefer it. I'll go rename it.

PR #297, opened 2025-09-05
feat(typeEvaluator): support document::get -> unknown/null

sgulseth:
### Description

Adds support for parsing `documents::get` so it doesn't break type evaluation


tzhelyazkova (approved):
looks ok to me

PR #314, opened 2025-11-20
fix(typeEvaluator): handle splatting arrays

sgulseth:
### Description

We didn't handle the case where an array was splatting an array element, this lead to `[...[1]]` being typed as `Array<Array<1>>` instead of `Array<1>`

It should now match the evaluator in https://github.com/sanity-io/groq-js/blob/main/src/evaluator/evaluate.ts#L606-L611

### Testing

Added


kbrabrand (approved):
LGTM

PR #316, opened 2025-12-08
fix(evaluator): handle numerical addition/subtraction from datetime

kbrabrand:
### Description

Adds support for add/subtract operations where the left operand is a datetime string and the right operand is a number. If a valid datetime string is passed the type is a string, while for any invalid datatime values the type is null. Based on this we return a `string-null` union.

This fixes the underlying issue reported in sanity-io/sanity#8255 where the generated type for the following query would not be correct:

```
(dateTime("2025-03-01T00:00:00Z") + 60) > dateTime("2024-03-01T00:00:00Z")
```

it now yields a `boolean-null` union, as is correct since an invalid dateTime string would result in `null` any comparison operation that includes a `null` operand evaluates to `null`.

https://linear.app/sanity/issue/CLDX-3609/groq-js-support-datetime-numerical-operations-in-type-evaluator

### What to review

The code

### Testing

Have tested the whole typegen flow in a local project with both a simple subtraction and a subtraction as part of a comparison operation. Also added a test for the case that failed before, with `dateTime(...) - number`.

### DIFF
--- a/src/typeEvaluator/functions.ts
+++ b/src/typeEvaluator/functions.ts
@@ -4,7 +4,7 @@ import {optimizeUnions} from './optimizations'
 import type {Scope} from './scope'
 import {walk} from './typeEvaluate'
 import {createGeoJson, mapNode, nullUnion} from './typeHelpers'
-import type {NullTypeNode, TypeNode} from './types'
+import {STRING_TYPE_DATETIME, type NullTypeNode, type TypeNode} from './types'
 
 function unionWithoutNull(unionTypeNode: TypeNode): TypeNode {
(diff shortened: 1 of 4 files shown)


kbrabrand:
> Preferably these types of fixes should be done by adding support for dateTime in typeEvaluateCompare: https://github.com/sanity-io/groq-js/blob/df81e4a9f17bc71bfff44ca0e404586a55c47987/test/typeEvaluateCompare.test.ts#L71. This would give us far better coverage of how it works with all types of operators and that it's it sync with groq-js's implementation.

Cool. I've added a task for it in linear and will follow up later.

PR #317, opened 2025-12-08
feat(typeEvaluator): add support for number + datetime operation

kbrabrand:
### TL;DR

Adds support for commutative addition between numbers and datetime values.

https://linear.app/sanity/issue/CLDX-4468/add-support-for-number-datetime-in-groq-js

### What changed?

`number + datetime` is now supported in addition to `datetime + number`

### How to test?

Run the new test case that verifies a datetime can be added to a number:

```typescript
t.test('number + dateTime', (t) => {
  const query = `(60 + global::dateTime("2025-03-01T00:00:00Z")) > global::dateTime("2024-03-01T00:00:00Z")`
  const ast = parse(query)
  const res = typeEvaluate(ast, schemas)
  t.same(res, {
    type: 'union',
    of: [{type: 'boolean', value: undefined}, {type: 'null'}],
  })
  t.end()
})
```

### Why make this change?

Gradient supports these operations (even though they're not defined in the spec). We should update the spec too, but this change at least gets `groq-js` in line with the actual query engine behaviour, which is what users expect.

### DIFF
--- a/src/typeEvaluator/typeEvaluate.ts
+++ b/src/typeEvaluator/typeEvaluate.ts
@@ -34,7 +34,6 @@ import {optimizeUnions} from './optimizations'
 import {Context, Scope} from './scope'
 import {isFuncCall, mapNode, nullUnion, resolveInline} from './typeHelpers'
 import {
-  STRING_TYPE_DATETIME,
   type ArrayTypeNode,
   type BooleanTypeNode,
   type Document,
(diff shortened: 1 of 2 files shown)


kbrabrand:
### Merge activity

* **Dec 9, 9:20 AM UTC**: @kbrabrand merged this pull request with [Graphite](https://app.graphite.com/github/pr/sanity-io/groq-js/317).

PR #344, opened 2026-04-14
refactor: separate parse from evaluate for bundle size

rexxars:
## Summary

- Decouples the parser from the evaluator so consumers who only use `parse()` can tree-shake away the evaluator code
- Introduces a lightweight function metadata registry (`src/functionRegistry.ts`) that the parser uses for validation instead of importing full function implementations
- Extracts a standalone constant evaluator (`src/constantEvaluate.ts`) that handles parse-time syntax disambiguation without depending on the evaluator
- Configures the build to split output into separate chunks (shared + evaluator) so bundlers can eliminate evaluator code

## Motivation

Previously, importing `parse` pulled in the entire evaluator because:
1. The parser imported `namespaces` and `pipeFunctions` from `evaluator/functions/` to validate function calls, which transitively imported every function implementation and the core evaluator
2. The parser imported `tryConstantEvaluate` from `evaluator/`, which depended on `executeSync`
3. `FuncCallNode` and `PipeFuncCallNode` carried runtime function references (`func: GroqFunction`), creating a type-level dependency from `nodeTypes.ts` to the evaluator

After this change, the parser only depends on pure metadata (function names, arity, mode) and a standalone constant evaluator - both with zero evaluator imports.

## Bundle size impact

Measured with rollup + `@rollup/plugin-node-resolve` (respects `sideEffects: false`):

- **Parse-only: ~63 KB** (parser, raw parser, type evaluator, values, function registry)
- **Full (parse + evaluate): ~114 KB**
- **Reduction: ~45%** when only using `parse()`

## Breaking changes

**Removed properties from AST node types:**

- `FuncCallNode.func` (`GroqFunction`) - removed, use `namespace` + `name` to identify the function
- `PipeFuncCallNode.func` (`GroqPipeFunction`) - removed, use `name` to identify the function

These were runtime function references embedded in the AST at parse time. Consumers inspecting parsed AST nodes for `.func` will see `undefined`. The `namespace`, `name`, and `args` fields are unchanged.

I'm not sure if we should _release_ this as a breaking change - codegen does not appear to be using these, and I'd find it surprising if anyone _was_ using these for any purpose - but would be interested in hearing your opinion on that.

## How it works

1. `src/functionRegistry.ts` contains all function metadata (name, arity, mode) as plain data - no imports
2. The parser validates function calls against the registry and builds AST nodes with `namespace` + `name` (no `func` reference)
3. At evaluation time, the evaluator looks up function implementations by `namespace` + `name` from the existing `namespaces`/`pipeFunctions` objects
4. The build config (`package.config.ts`) uses `manualChunks` to split evaluator code into a separate chunk that bundlers can eliminate


package.config.ts
```diff
@@ -1,10 +1,55 @@
...
 export default defineConfig({
   extract: {
     rules: {
       'ae-internal-missing-underscore': 'off',
       'ae-missing-release-tag': 'warn',
     },
   },
+  rollup: {
+    output: {
+      manualChunks(id) {
```
judofyr, line 16: This logic is not tested. Can we add a test? Maybe just a custom script (`npm run test-bundle`) which runs `esbuild` with bundle and checks which identifiers are _not_ part of the bundle. And then just run it in GitHub Action.

package.config.ts
```diff
@@ -1,10 +1,55 @@
...
+        if (
+          isUnder(relative, 'values') ||
+          name === 'nodeTypes' ||
+          name === 'functionRegistry' ||
+          name === 'constantEvaluate'
+        ) {
+          return 'shared'
+        }
+
+        return undefined
```
judofyr, line 27: I'd rather do a restructure of the directory structure based on this, so that these rules can be much simpler: E.g. Introduce a `src/shared/` directory and move all the files that are returned as `shared` here to it.

That way the manual chunks can just be:

- `src/shared` => `shared`
- `src/evaluator` => `evaluator`

package.config.ts
```diff
@@ -1,10 +1,55 @@
...
+
+        // evaluator/matching is also used by the type evaluator, so it
+        // must be in the shared chunk to avoid a circular dependency.
+        if (dir === 'evaluator' && name === 'matching') {
+          return 'shared'
+        }
+
+        if (isUnder(relative, 'evaluator') && name !== 'types') {
+          return 'evaluator'
+        }
```
judofyr:
```suggestion
        if (isUnder(relative, 'evaluator')) {
          return 'evaluator'
        }
```

I couldn't see any difference by removing this `name !== 'types'` since it only contains TS definition.

judofyr (changes requested):
- +1 on the change with the function registry. Looks good!
- Blocker: We need some tests for this bundling behavior unless it _is_ going to break very soon.
- Non-blocker: I'm also very much open for restructuring the repo so that the rules around chunking is simpler. That also makes it less fragile in the future. Maybe even enforcing it with https://github.com/javierbrea/eslint-plugin-boundaries? Could be done in follow-up PR as well.

package.config.ts
rexxars: Test added

package.config.ts
rexxars: Agree!

judofyr (approved):
> +3,206  -2,238

Now this diff is way too big 🤪

jk jk. Looks good!

PR #349, opened 2026-06-05
ci: add pkg.pr.new continuous releases

rexxars:
Sets up [pkg.pr.new](https://github.com/stackblitz-labs/pkg.pr.new) so every pull request and every push to `main`/`canary` publishes an instant preview release, installable from a pkg.pr.new URL without publishing anything to npm.

## What's included

- **`.github/workflows/pkg-pr-new.yml`** - clean install, `npm run build`, then `npm exec pkg-pr-new publish`. The build runs first because `dist` must exist before publishing: `npm pack` does not run the `prepublishOnly` hook.
- **`package.json`** / **`package-lock.json`** - `pkg-pr-new` added as a devDependency. Per their CI guidance, it's installed from the lockfile and run with `npm exec` rather than `npx`/`dlx`.

## Notes on the setup

- Triggers on `pull_request` plus pushes to `main`/`canary` - deliberately not bare `[push, pull_request]`, which would double-publish on every PR branch push.
- `permissions: {}` - PR comments and check runs come from the pkg.pr.new GitHub App, not the workflow token, so no extra permissions are needed.
- Compact URLs work out of the box since `groq-js` is on npm with a valid `repository` field, so previews install as `npm i https://pkg.pr.new/groq-js@<sha>`.

## Required before this works

The [pkg.pr.new GitHub App](https://github.com/apps/pkg-pr-new) must be installed on `sanity-io/groq-js`. It needs to be installed before the first publish runs, or the workflow step will fail.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -22,4 +22,4 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npm run build
-      - run: npm exec pkg-pr-new publish
+      - run: npm exec pkg-pr-new publish --no-template
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -9,7 +9,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
-permissions: {}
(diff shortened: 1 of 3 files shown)


binoy14 (approved):
LGTM

